PR #221, opened 2023-01-09
chore: add code formatter and linter to the codebase

marieglr:
## What

This PR adds Prettier for code formatting and reconfigured ESLint for code style rules. 

- ESLint was already present, this PR merges config files that were located in 2 places (in package.json, and inside the test folder), and harmonises rules with rules defined in the Xemx codebase. Feel free to comment and make requests about any fomatting or style rules you would like us to agree on. 
- Finally, this PR adds a lint step in the CI. 

## What to check
- .eslintrc.js holds the eslint config
- .prettierrc holds the prttier config
- .eslintignore holds a list of files not to run linters on
- .circleci/config.yml for the ci step

Most changes on other files were made automatically, but you can check the commit history for manual fixes. 
Main two offenses to fix manually were:
- [no shadow](https://eslint.org/docs/latest/rules/no-shadow)
- r[equire-default-prop](https://eslint.vuejs.org/rules/require-default-prop.html) --> I think this one is a really good rule, but I ended up turning it off, because it was too much refactor to implement it. I tried marking the props required or adding defaults where I could, but tests were failing and there is about 100 occurrences to fix, which is too much without prior knowledge of the codebase. 

## How to use
To run ESLint in autocorrect mode, run `yarn lint:fix`, this will also autocorrect prettier offences. 
You can also configure your own editor tu run this automatically. 

## Why
fixes https://github.com/ICIJ/datashare/issues/977 and https://github.com/ICIJ/datashare/issues/976

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,7 +3,7 @@ module.exports = {
     jest: true,
     node: true
   },
-  extends: ['plugin:vue/strongly-recommended', 'plugin:vue/recommended', 'standard', 'prettier'],
+  extends: ['plugin:vue/recommended', 'standard', 'prettier'],
   plugins: ['prettier'],
   rules: {
     'prettier/prettier': 'error',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,20 @@
+module.exports = {
+  env: {
+    jest: true,
+    node: true
+  },
+  extends: ['plugin:vue/strongly-recommended', 'standard', 'plugin:vue/essential', '@vue/standard', 'prettier'],
+  plugins: ['prettier'],
+  rules: {
+    'prettier/prettier': 'error',
+    'import/no-webpack-loader-syntax': 'off',
+    'no-useless-escape': 'off',
+    'lines-between-class-members': 'off',
+    'template-curly-spacing': 'off',
+    'vue/custom-event-name-casing': 'off',
+    'import/no-extraneous-dependencies': 'off'
+  },
+  parserOptions: {
+    parser: 'babel-eslint'
+  }
+}
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,7 +3,7 @@ module.exports = {
     jest: true,
     node: true
   },
-  extends: ['plugin:vue/strongly-recommended', 'standard', 'prettier'],
+  extends: ['plugin:vue/strongly-recommended', 'plugin:vue/recommended', 'standard', 'prettier'],
   plugins: ['prettier'],
   rules: {
     'prettier/prettier': 'error',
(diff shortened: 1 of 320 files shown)


.eslintrc.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
```
pirhoo, line 1: Any reason why this was moved out of the package.json? I don't mind, I'm just curious :)

marieglr: There was a `.eslintrc.js` in the `tests` forlder, I thought I would merge both for clarity (as it is not obvious that a `.eslintrc` file and a `eslintConfig` object in the package.json are exactly the same thing). And also to lighten the package.json and separate concern, but I don't mind either putting it back :) 

Your comment made me realise I didn't remove the second .eslintrc, doing it now!

pirhoo (approved):

.eslintrc.js
```diff
@@ -0,0 +1,18 @@
...
+  env: {
+    jest: true,
```
caro3801, line 3: related to file name: since we have .prettierrc shouldn't we have have .eslintrc instead of .eslintrc.js (or the other way around) ?

src/components/BatchSearchFilterDropdown.vue
```diff
@@ -1,7 +1,13 @@
 <template>
   <batch-search-filter class="batch-search-filter-dropdown" :id="id" :name="name" :active="isActive">
     <keep-alive>
-      <selectable-dropdown v-model="selectedValues" :items="items" deactivate-keys :multiple="multiple" class="shadow-none border-0">
+      <selectable-dropdown
+        v-model="selectedValues"
```
caro3801, line 5: I think we can add the vue guidelines to have order in the props (enhance readability from my POV):
- v-if
- bindings
- alphabetical order

caro3801 (approved):
:+1: :muscle:

src/components/BatchSearchFilterDropdown.vue
marieglr: Ah yes,  that'd be nice! do you know where I can find these as eslint rules?

caro3801: It's from the plugin eslint vue https://eslint.vuejs.org/
https://eslint.vuejs.org/rules/attributes-order.html
my bad I think it's already enforced...

mvanzalu (approved):

.eslintrc.js
```diff
@@ -3,7 +3,7 @@ module.exports = {
     jest: true,
     node: true
   },
-  extends: ['plugin:vue/strongly-recommended', 'standard', 'prettier'],
+  extends: ['plugin:vue/strongly-recommended', 'plugin:vue/recommended', 'standard', 'prettier'],
```
caro3801, line 6: I think you can remove 'plugin:vue/strongly-recommended' : it is overrided by 'plugin:vue/recommended' and 'plugin:vue/recommended' contains the rules of 'plugin:vue/strongly-recommended'
https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/configs/vue3-recommended.js

bamthomas (approved):

PR #222, opened 2023-01-10
refactor: remove offline doc

pirhoo:
## Motivation

Currently, Datashare's user guide is shipped with the client and can be browsed offline:

Screen Shot 2023-01-10 at 19 30 05

Contextual links to the user guide are displayed in the menu:

Screenshot from 2023-01-10 19-31-46

But this feature comes with a huge cost: **the final deliverable of Datashare is 130MB heavier**.

To reduce the size of the application and simplify maintenance, we decided to remove the offline version from Datashare. We still want to display contextual links to the user guide, leading directly to Gitbook.

### Routes definition

The most important changes was done in the routes definition files, where user guide pages were define using the path to their markdown file.

It was replace by a more transparent definition object, with 3 properties:

* `path`: relative path to the user guide page (without the Githbook host) ;
* `title`: title of the page, which is plain English for now but it can be a translation key ;
* `mode` (optional): a string or an array of string listing the mode on which this link is visible.

For instance:

```js
{
  name: 'landing',
  path: '',
  component: () => import('@/pages/Landing'),
  meta: {
    docs: [
      {
        title: 'Add documents to Datashare',
        path: '<%- os %>/add-documents-to-datashare-on-<%- os %>',
        mode: ['LOCAL', 'EMBEDDED']
      },
      {
        title: 'Analyse documents',
        path: 'all/analyze-documents'
      }
    ]
  }
}
```

### Contextual links

The list of contextual links is still built in a mixin, so it can be used by different components (in case we want to move it out of the `AppSidebar`).

This mixin uses to build a computed property called `currentRouteDocs`. It was replaced by a new computed property called `filteredRouteDocs` which simply:

1. reads the current route's metadata
2. treats the `path` as a template string to inject the current OS
3. try to find a translation key for the given title
4. filter the list of links according to the current app mode (LOCAL, SERVER, etc).

## Post-face

For the sack of readability I did not move the "public/docs" to the root of the project. This mean it will still be shipped with Datashare. I intend to make the move directly on master and GitBook after the branch has been merged in master.


marieglr (approved):
Very clear explanations, thank you! 
Also tested locally, works like a charm.
One small question: do we not need to test this part?

pirhoo:
> One small question: do we not need to test this part?

We might! But that's an other refactoring :)

bamthomas (approved):

PR #225, opened 2023-01-12
fix: language ignored in modals

pirhoo:
This PR aims at fixing the persistence of the selected locale.

I used this opportunity to move the logic of loading/settings the current locale into a proper core mixin called `I18nMixin`. This mixin is now used by the `LocalesMenu` component and by the `Core` instance when configuring the app.

The non-persisting state was due to the fact that the current locale was loaded directly within the `LocalesMenu` component at mount time. So for routes where there was no `LocalesMenu` (like the modal, or the login screen), the locale was never loaded. It is now done at the `Core` level, so for every route.

### DIFF
--- a/tests/unit/specs/components/LocalesMenu.spec.js
+++ b/tests/unit/specs/components/LocalesMenu.spec.js
@@ -5,30 +5,30 @@ import LocalesMenu from '@/components/LocalesMenu'
 import { Core } from '@/core'
 
 describe('LocalesMenu', () => {
-  const { localVue, i18n } = Core.init(createLocalVue()).useAll()
+  const core = Core.init(createLocalVue()).useAll()
+  const { localVue, i18n } = core
   let wrapper = null
 
-  afterEach(() => {
-    localStorage.removeItem('locale')
-    i18n.locale = 'en'
+  beforeEach(async () => {
+    await core.loadI18Locale('en')
   })
 
-  describe('should change the interface language according to localStorage', () => {
+  describe('should change the interface language according to configuration', () => {
     it('should display the interfaces in English by default', () => {
       wrapper = shallowMount(LocalesMenu, { localVue, i18n })
 
       expect(wrapper.find('.locales-menu__button').text()).toBe('English')
     })
 
-    it('should display the interface in French if localStorage says so', () => {
-      localStorage.setItem('locale', 'fr')
+    it('should display the interface in French if configuration says so', () => {
+      core.setI18nLocale('fr')
       wrapper = shallowMount(LocalesMenu, { localVue, i18n })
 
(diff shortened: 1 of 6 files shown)


tests/unit/specs/core/I18nMixin.spec.js
```diff
@@ -0,0 +1,84 @@
...
+    it('should have "de" messages', () => {
+      expect(core.hasI18Locale('de')).toBeTruthy()
+    })
+
+    it('should switch to Japanese (ja)', async () => {
+      await core.loadI18Locale('ja')
+      expect(core.i18n.locale).toBe('ja')
+    })
+
+    it('should have "ja" mjasage', async () => {
```
caro3801, line 74: typo ? "mjasage" => "messages"

pirhoo: Oupsy

tests/unit/specs/core/I18nMixin.spec.js
```diff
@@ -0,0 +1,84 @@
...
+    beforeEach(async () => {
+      localStorage.removeItem('locale')
+      core = Core.init(createLocalVue()).useAll()
+      await core.initializeI18n()
+    })
+
+    afterEach(async () => {
+      await core.loadI18Locale('en')
+    })
+
```
caro3801, line 17: I would add a afterAll hooh with `localStorage.removeItem('locale')`

tests/unit/specs/components/LocalesMenu.spec.js
```diff
@@ -5,30 +5,30 @@ import LocalesMenu from '@/components/LocalesMenu'
...
+  const core = Core.init(createLocalVue()).useAll()
+  const { localVue, i18n } = core
   let wrapper = null
 
-  afterEach(() => {
-    localStorage.removeItem('locale')
-    i18n.locale = 'en'
+  beforeEach(async () => {
+    await core.loadI18Locale('en')
   })
```
caro3801, line 15: Maybe adding an afterAll with `await core.loadI18Locale('en')` would enforce a teardown without side effects on the locale.

tests/unit/specs/core/I18nMixin.spec.js
pirhoo: Good call!

tests/unit/specs/components/LocalesMenu.spec.js
pirhoo: Good call too!

caro3801 (approved):
Nice refactor !

pirhoo:
Thanks @caro3801! I've made the suggested changes

PR #226, opened 2023-01-13
aria: remove sliding transition

pirhoo:
Simple accessibility improvement removing a transition when opening a document.


marieglr (approved):
Didn't know this kind of very short animation was damaging accessibility!

PR #230, opened 2023-01-17
fix: filter button color

marieglr:
**Before this PR,** when two SearchFormControl were displayed on the page at the same time, the color of one would overwrite the one of the other. 
More specifically, when opening the "Named Entities" tab on a document, the search buttons on the filters in the left panel would become white, as they share the same component as the search button in the named entities tab. 

**After this PR,** the css to display the button's color is delegated to the SearchFormControl itself to avoid collusions.


src/components/filter/FilterBoilerplate.vue
```diff
@@ -34,7 +33,8 @@
             v-model="query"
             class="filter__items__search"
             :placeholder="$t('search.searchIn') + ' ' + $t('filter.' + filter.name) + '...'"
-            :rounded="false"
+            rounded="false"
+            dark="true"
```
pirhoo, line 37: Boolean prop doesn't need a value when they have a `false` as default value.

In that case you also pass "false" (as a string) to the rounded prop which expects a boolean.

So the properties should be written:

```
:rounded="false"
dark
```

pirhoo: Note that it's not very intuitive to have positive default value for `autofocus` and `rounded` on this component. We might refactor their default values to `false` and update the place where this component is used.

src/components/filter/FilterBoilerplate.vue
```diff
@@ -213,13 +213,6 @@ export default {
       type: Boolean,
       default: true
     },
-    /**
-     * Display the filter on dark background
-     */
-    dark: {
```
pirhoo, line 219: We should keep the `dark` property on this component (with the CSS you already removed) but pass this property down to `SearchFormControl`. This way the filter can also be display in its modal (non dark).

So the properties should be written:

```
:rounded="false"
:dark="dark"
```

marieglr: Alright thanks, I was wondering if this was still used, I'll add it back

src/components/SearchFormControl.vue
```diff
@@ -63,6 +63,11 @@ export default {
     loading: {
       type: Boolean,
       default: false
+    },
+
```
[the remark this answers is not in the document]
marieglr: thanks for the tips!

src/components/filter/FilterBoilerplate.vue
```diff
@@ -642,21 +635,6 @@ export default {
...
-      .search-form-control__addon__submit:last-of-type {
-        background: #000;
-        color: inherit;
-      }
-
-      .search-form-control__input:not(:focus),
-      .search-form-control__addon__submit:last-of-type {
-        border-color: #000;
-      }
-    }
```
marieglr, line 658: @pirhoo to be clear, I should also add this back? won't it overwrite the css in the subcomponent?

pirhoo: Nop, this part is now handle by the search-form-control component and triggered by the dark property you pass down.

marieglr: okay, I misunderstood, sorry

pirhoo (approved):

PR #231, opened 2023-01-17
refactor: add two inputs for date range filter

caro3801:
This PR is related to ICIJ/datashare#981

Added 2 inputs for start/end date range picking. The date format changes according to the locale. 
UX: 
- title / alt / placeholder for helping on date format
- when fields are empty and start is filled, the ui focuses on the end field.
Fix: 
- date filters where not loaded on page load 

Screenshots:
Screenshot from 2023-01-17 15-14-49

Screenshot from 2023-01-17 15-13-27

### DIFF
--- a/src/components/filter/types/FilterDateRange.vue
+++ b/src/components/filter/types/FilterDateRange.vue
@@ -68,8 +68,8 @@ export default {
         if (values.length < 2) {
           return null
         }
-        const start = min(values)
-        const end = max(values)
+        const start = parseInt(min(values))
+        const end = parseInt(max(values))
         return { start, end }
       },
       set(range) {
--- a/src/components/filter/types/FilterDateRange.vue
+++ b/src/components/filter/types/FilterDateRange.vue
@@ -81,9 +81,12 @@ export default {
       return [
         {
           key: 'today',
-          contentStyle: {
-            fontWeight: '700',
-            fontSize: '.9 rem'
+          highlight: {
+            style: {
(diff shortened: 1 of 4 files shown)


pirhoo:
Great! Can we also ensure it's the same colors (at least the background in gray dark) please?

src/components/filter/types/FilterDateRange.vue
```diff
@@ -46,6 +71,11 @@ export default {
     FilterBoilerplate
   },
   extends: FilterAbstract,
+  data() {
+    return {
+      placeholderMask: ''
```
pirhoo, line 76: Any reason why `placeholderMask` is not a computed property?

caro3801: yes it's because it's based on a ref and from what I understood refs can't be "computed" that's why I reassign it on locale change: https://github.com/ICIJ/datashare-client/blob/d2c3747ace755bb6e7e550b4101370f887911e9f/src/components/filter/types/FilterDateRange.vue#L129

pirhoo: Makes sense! Thanks

caro3801:
Adding color theme to the popover in 0cda3040de57100d54a76f6b4a5d63dc1de4e47b (related to ICIJ/datashare#981). 
The 18's highlight is because it's today 
image

pirhoo (approved):

PR #233, opened 2023-01-19
feat: add batch search projects multiselect

caro3801:
Related to ICIJ/datashare#992

Add [vue-multiselect](https://vue-multiselect.js.org/#sub-getting-started) with same [style](https://github.com/ICIJ/prophecies/blob/main/prophecies/apps/frontend/src/utils/_multiselect.scss) as done on ICIJ/prophecies => proposal for future steps could be to add it to murmur.

Screenshot from 2023-01-19 12-00-19

Screenshot from 2023-01-19 12-00-45

### DIFF
--- a/src/components/BatchSearchForm.vue
+++ b/src/components/BatchSearchForm.vue
@@ -4,21 +4,25 @@
       <h5 v-if="!hideTitle" class="py-2 h6 text-uppercase text-muted">
         {{ $t('batchSearch.heading') }}
       </h5>
-      <div class="card w-100" :class="{ 'border-0': hideBorder }">
+      <div :class="{ 'border-0': hideBorder }" class="card w-100">
         <div class="card-body pb-1">
-          <b-form-group label-size="sm" :label="`${$t('batchSearch.name')} *`">
-            <b-form-input v-model="name" type="text" required></b-form-input>
+          <b-form-group :label="`${$t('batchSearch.name')} *`" class="batch-search-form__name" label-size="sm">
+            <b-form-input v-model="name" required type="text"></b-form-input>
           </b-form-group>
-          <b-form-group label-size="sm" class="mb-0" :label="`${$t('batchSearch.fileLabel')} *`">
+          <b-form-group
+            :label="`${$t('batchSearch.fileLabel')} *`"
+            class="batch-search-form__fileLabel mb-0"
+            label-size="sm"
+          >
             <template slot="description">
               <div v-html="$t('batchSearch.fileDescription')"></div>
             </template>
             <b-form-file
               v-model="csvFile"
               :placeholder="$t('batchSearch.filePlaceholder')"
+              :state="Boolean(csvFile)"
               accept=".csv"
               class="text-truncate"
-              :state="Boolean(csvFile)"
               no-drop
               required
             ></b-form-file>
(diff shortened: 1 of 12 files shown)


src/components/BatchSearchForm.vue
```diff
@@ -4,21 +4,25 @@
...
+            class="batch-search-form__fileLabel mb-0"
+            label-size="sm"
+          >
             <template slot="description">
               <div v-html="$t('batchSearch.fileDescription')"></div>
             </template>
             <b-form-file
               v-model="csvFile"
               :placeholder="$t('batchSearch.filePlaceholder')"
+              :state="Boolean(csvFile)"
```
pirhoo, line 23: I know it was already there but this state should be a computed property. I would have preferred the shorthand `!!` too.

pirhoo, line 23: I know it was already there but this state should be a computed property. I would have preferred the shorthand `!!` too.

src/components/BatchSearchForm.vue
```diff
@@ -32,28 +36,28 @@
...
+            :label="`${$t('batchSearch.projects')} *`"
+            class="batch-search-form__projects"
             label-size="sm"
+          >
+            <multiselect
+              v-model="projects"
+              :allow-empty="false"
+              :multiple="true"
+              :options="availableProjects"
+              :searchable="true"
```
pirhoo, line 53: The props `:multiple="true"` and `:searchable="true"` can be set without `="true"`. Also `searchable` is default to true so we can remove it.

src/components/BatchSearchForm.vue
```diff
@@ -342,12 +358,23 @@ export default {
...
+          each(aggTypes, (aggType) => {
+            const extensions = has(types, aggType) ? types[aggType].extensions : []
+            const label = has(types, aggType) ? types[aggType].label : aggType
+            this.allFileTypes.push({
+              extensions,
+              label,
+              mime: aggType
+            })
+          })
+        } catch (e) {
```
pirhoo, line 372: Good call catching this error

src/components/BatchSearchForm.vue
```diff
@@ -373,6 +400,7 @@ export default {
     },
     async onSubmit() {
       try {
+        if (!this.projects?.length) return
```
pirhoo, line 403: Is this necessary? If we have the `allow-blank=false` on the multiselect, couldn't we use the form validation to disable the submit button?

pirhoo (approved):
Excellent addition, very well done! Thanks @caro3801!

mvanzalu (approved):

src/utils/_multiselect.scss
```diff
@@ -0,0 +1,115 @@
+@import url('~node_modules/vue-multiselect/dist/vue-multiselect.min.css');
```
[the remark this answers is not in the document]
pirhoo: It is shipped with vue-multiselect yes

src/components/BatchSearchForm.vue
caro3801: Me too; it's changed

src/components/BatchSearchForm.vue
caro3801: :broom:

src/utils/_multiselect.scss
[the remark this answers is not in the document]
caro3801: The `vue-multiselect.min.css`  is the original css file from the lib that has been minified (automatically). 
In order to adapt the component style to our project, we added `src/utils/_multiselect.scss` which import the orignal css in order to override its properties.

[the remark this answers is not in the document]
mvanzalu: Got it ! THanks

Issue #234, opened 2023-01-23
bug: batch search publishState is not reflected in the filter modal


caro3801:
Current behavior

![Image](https://user-images.githubusercontent.com/4643139/214096642-40ecd9dc-3bd2-4bee-9b05-39f93175b158.png)

Wanted: "Published" should be highlighted

PR #235, opened 2023-01-24
bug: add published dropdown filter selected on publish state update #234

caro3801:
related to #234 
before:
publicationStatus dropdown list item was not highlighted when updated (selected/unselected)

after:
publicationStatus dropdown list item is highlighted when updated

why: we use the same dropdown for single value selection and multiple selection. the single value is an object with label/value therefore we need to handle the selected callback differently

:v: add test on publication status router update

### DIFF
--- a/src/components/BatchSearchFilterDropdown.vue
+++ b/src/components/BatchSearchFilterDropdown.vue
@@ -6,6 +6,7 @@
         :items="items"
         deactivate-keys
         :multiple="multiple"
+        :eq="eq"
         class="shadow-none border-0"
       >
         <template #item-label="{ item }">
@@ -19,6 +20,7 @@
 <script>
 import BatchSearchFilter from '@/components/BatchSearchFilter'
 import { isEqual } from 'lodash'
+import eq from 'lodash/eq'
 
 export default {
   name: 'BatchSearchFilterDropdown',
@@ -31,6 +33,10 @@ export default {
     id: {
       type: String,
       required: true
+    },
+    eq: {
+      type: Function,
+      default: eq
     },
     name: {
       type: String,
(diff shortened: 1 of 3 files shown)


src/components/BatchSearchFilterDropdown.vue
```diff
@@ -32,6 +34,10 @@ export default {
       type: String,
       required: true
     },
+    eq: {
+      type: Function,
```
pirhoo, line 38: Why did you pass a function instead of a boolean here?

bamthomas (approved):

src/components/BatchSearchTable.vue
```diff
@@ -445,7 +451,9 @@ export default {
     getProjectsNames(item) {
       return item.projects?.map((project) => project.name).join(', ') ?? ''
     },
-
+    isSelected(item) {
```
pirhoo, line 454: If you switch to a boolean prop for `eq`, the method `isSelected` should be a computed property.

src/components/BatchSearchFilterDropdown.vue
caro3801: It's related to the behavior of selectable-dropdown in murmur:
https://github.com/ICIJ/murmur/blob/a9a77a00e36811d8a0e10cfb2754506cda6a2774/lib/components/SelectableDropdown.vue#L112-L118

The equality function is relative to one item at a time that's why it passes down the function as a prop. I don't see how it can be computed in advance on a upper level.

pirhoo: Ah, yes, Murmur! I understand now. Makes more sens, thanks!

pirhoo (approved):

PR #236, opened 2023-01-24
Fix/extract missing lang

caro3801:
Related to ICIJ/datashare#995

before:
image

After:
- No OCR 
image

- Error on OCR retrieve => submit is disabled when OCR is on
image

- OCR + ocr languages available

image

image


Soliine:
It looks great!

bamthomas (approved):

PR #240, opened 2023-02-14
refactor: update nbBatchSearches when submitting new BS #1019

bamthomas:
This refactor is related to :
1. ICIJ/datashare#1019 : No error on delete operation
2. and https://github.com/ICIJ/datashare-client/pull/239 to improve clarification of total vs nbBatchSearches : `total` is the total of BS with filters on (updated when getting batch searches with filters), `nbBatchSearches` is the number of Batch searches without filtering.  

`nbBatchSearches` is updated: 
- when mounting the component, 
- on submit of a new BS,
- and on delete.

Co-authored-by: Caroline Desprat <caro3801@users.noreply.github.com>


pirhoo (approved):

src/pages/BatchSearch.vue
```diff
@@ -34,23 +34,23 @@ export default {
...
     },
     noBatchSearch() {
       return this.$t('batchSearch.empty', { howToLink: this.howToLink })
     }
   },
-  async mounted() {
+  mounted() {
     if (!this.hasBatchSearch) {
-      await this.getBatchSearch()
+      this.getBatchSearches()
```
mvanzalu, line 47: `getBatchSearches()` is async right ?

mvanzalu (approved):

src/pages/BatchSearch.vue
caro3801: In this case the promise does not need to be "awaited" since it's in the `mounted` hook and there is nothing to do after that requires to await the result of the promise.

PR #241, opened 2023-02-15
feat: use pathSeparator from server settings ICIJ/datashare#1012

mvanzalu:


src/components/InlineDirectoryPicker.vue
```diff
@@ -22,12 +22,13 @@ export default {
     return {
       browse: false,
       browsingPath: null,
-      browsingTree: null
+      browsingTree: null,
+      separator: this.$core.config.get('pathSeparator')
```
pirhoo, line 26: I would have used the same name for this property in the data object (`pathSeparator`) to make it consistent.

pirhoo (approved):

src/components/InlineDirectoryPicker.vue
mvanzalu: Yes indeed, fixing this

pirhoo (approved):

PR #242, opened 2023-02-16
fix: correct TreeBreadcrumb root on Windows ICIJ/datashare#1012

mvanzalu:


src/components/TreeBreadcrumb.vue
```diff
@@ -74,13 +74,23 @@ export default {
       default: false
     }
   },
+  data() {
+    return {
+      pathSeparator: this.$core.config.get('pathSeparator')
```
caro3801, line 79: What happen if the pathSeparator is not configured (or wrongly defined)? do we have a fallback?

caro3801 (changes requested):
Adding a test with a different (windows?) separator to document the fix would be a great addition

src/components/TreeBreadcrumb.vue
pirhoo: `$config.get` accepts a default value: `this.$core.config.get('pathSeparator', '/')`

mvanzalu: Good remark

caro3801 (approved):

PR #244, opened 2023-02-20
fix: ensure store is reseted when URL query change

pirhoo:
The error described in https://github.com/ICIJ/datashare/issues/1022 is caused by the method that read search filter from the URL.

This PR takes care of resetting the search filter every time an attempt to read route parameters is done.


marieglr (approved):
nice catch and well tested, interesting for me :+1:

pirhoo:
Thanks @marieglr !

PR #245, opened 2023-02-24
feat: add links to project names (ICIJ/datashare#1018)

marieglr:
fixes https://github.com/ICIJ/datashare/issues/1018

This PR makes project  names clickable in batch searches, with a link pointing to searching in the project. 

image

image


pirhoo:
Thanks! Just one change: if the project is clickable I would have move it to the end or the middle column (usually the first clickable column is implicitly a link to the record).

src/pages/BatchSearchResults.vue
```diff
@@ -551,6 +558,9 @@ export default {
       this.$set(this, 'documentInModalPageIndex', pageIndex)
       this.$bvModal.show('document-modal')
     },
+    projectLink(projectName) {
+      return `/#/?indices=${projectName}`
```
pirhoo, line 562: It's better to use a route resolve function:

```
const query = { indices: projectName }
return this.$router.resolve({ name: 'search', query })
```

pirhoo (changes requested):

src/components/BatchSearchTable.vue
```diff
@@ -96,7 +96,7 @@
       </template>
       <template #cell(projects)="{ item }">
         <span v-b-tooltip.hover class="batch-search-table__item__projects text-truncate" :title="item.projectNames">
-          {{ item.projectNames }}
+          <a :href="projectLink(item.projectNames)"> {{ item.projectNames }}</a>
```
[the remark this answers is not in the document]
marieglr: nice, thanks both!

pirhoo (approved):

src/pages/BatchSearchResults.vue
```diff
@@ -24,7 +24,20 @@
...
+                  :to="{
+                    name: 'search',
+                    query: {
+                      q: '*',
+                      indices: project.name
+                    }
+                  }"
+                  class="batch-search-results__info__project-link"
+                >
+                  {{ project.name }}<span v-if="isNotLastArrayItem(index, batchSearch.projects.length)">,</span>
```
caro3801, line 39: I think it would be more pertinent  if the comma is outside of the link unless it's looking weird

caro3801 (approved):
Good improvements thanks !

src/pages/BatchSearchResults.vue
marieglr: Good idea, I changed it

PR #247, opened 2023-02-27
feat: add overlay spinner while submitting the form ICIJ/datashare#996

caro3801:
Related to https://github.com/ICIJ/datashare/issues/996 to avoid the frozen effect of the form on submit

image


src/components/ExtractingForm.vue
```diff
@@ -91,14 +94,21 @@ export default {
...
+    isWaitingForSubmitExtract() {
+      return this.$wait.is('submitExtract')
+    }
   },
   methods: {
+    dispatchExtract: waitFor('submitExtract', function () {
+      return this.$store.dispatch('indexing/submitExtract')
+    }),
     async submitExtract() {
       this.disabled = true
```
pirhoo, line 107: Now we have a waiter, disabled can be turned into a computed property:

```js
disabled () {
  return this.$wait.is('submitExtract')
}
```

PR #250, opened 2023-02-28
fix(history/search): display one badge per filter value

marieglr:
fixes https://github.com/ICIJ/datashare/issues/1032

Before This PR, when a filter had several value, the history for this search would display only one badge with one value.

After this PR, one badge per value for one filter is displayed (a filter with n values is displayed in n badges): 

image

### DIFF
--- a/src/pages/UserHistorySearch.vue
+++ b/src/pages/UserHistorySearch.vue
@@ -66,13 +66,10 @@ export default {
     },
     createFiltersFromURI(uri) {
       const urlSearchParams = new URLSearchParams(uri.split('?').slice(1).pop())
-      const params = Object.fromEntries(urlSearchParams.entries())
-      // Reduce params list into an array
-      return Object.keys(params).reduce((filters, name) => {
-        let value = params[name]
-        // Skip ignored param
+      const filters = []
+      for (let [name, value] of urlSearchParams.entries()) {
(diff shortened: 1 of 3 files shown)


src/pages/UserHistorySearch.vue
```diff
@@ -66,13 +66,10 @@ export default {
     },
     createFiltersFromURI(uri) {
       const urlSearchParams = new URLSearchParams(uri.split('?').slice(1).pop())
-      const params = Object.fromEntries(urlSearchParams.entries())
-      // Reduce params list into an array
-      return Object.keys(params).reduce((filters, name) => {
-        let value = params[name]
-        // Skip ignored param
+      const filters = []
+      for (let [name, value] of urlSearchParams.entries()) {
```
pirhoo, line 70: For the sake of simplicity, I would have used a `reduce`

pirhoo (approved):

src/pages/UserHistorySearch.vue
marieglr: we tried to but we could not use a `reduce` on `urlSearchParams.entries()` as it returns an iterator but not an array.
We had extensive discussions about readability of this method and were not completely satisfied, but this is a version we both accepted to yield too :relieved: if you want to try another version we can rediscuss it!

PR #253, opened 2023-03-06
feat: manage posix/win basenames in treeview ICIJ/datashare#1012

mvanzalu:
Related to [this issue](https://github.com/ICIJ/datashare/issues/1012).

On Windows, the basename of the folders isn't shown properly, instead, we have the full path : 
treebug  basename

This PR fixes the issue with this result : 
Treebug basename fixed

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "lucene": "^2.1.0",
     "moment": "^2.29.4",
     "mutationobserver-shim": "^0.3.7",
+    "path": "^0.12.7",
     "path-to-regexp": "^6.2.1",
     "release": "^6.3.1",
     "resize-observer-polyfill": "^1.5.1",
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "lucene": "^2.1.0",
     "moment": "^2.29.4",
     "mutationobserver-shim": "^0.3.7",
-    "path": "^0.12.7",
     "path-to-regexp": "^6.2.1",
     "release": "^6.3.1",
     "resize-observer-polyfill": "^1.5.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -44,6 +44,7 @@
     "lucene": "^2.1.0",
     "moment": "^2.29.4",
     "mutationobserver-shim": "^0.3.7",
+    "path": "^0.12.7",
```
pirhoo, line 47: I wouldn't use [path](https://github.com/jinder/path) since it's poorly maintained and not tested. You might want to use the Browserify variant instead: https://github.com/browserify/path-browserify

src/components/TreeView.vue
```diff
@@ -298,8 +299,8 @@ export default {
       const include = this.suffixPathTokens(this.pathSeparator.concat('.*')).join('|')
       const exclude = this.suffixPathTokens(this.pathSeparator.concat('.*', this.pathSeparator, '.*')).join('|')
       return {
-        include: this.usesWindowsSepator ? this.doubleWindowsSeparator(include) : include,
-        exclude: this.usesWindowsSepator ? this.doubleWindowsSeparator(exclude) : exclude,
+        include: this.usesWindowsSeparator ? this.doubleWindowsSeparator(include) : include,
+        exclude: this.usesWindowsSeparator ? this.doubleWindowsSeparator(exclude) : exclude,
```
pirhoo, line 303: I would have used two separated computed properties for `include` and `exclude` which themselves would use the `doubleWindowsSeparator` method if needed.

pirhoo (changes requested):

src/components/TreeView.vue
pirhoo: This change is cosmetic, I will approve the PR without it.

package.json
caro3801: Yeah but sadly "path-browserify only implements the POSIX functions, not the win32 ones."

PR #254, opened 2023-03-10
Refactor: use eslint-cong-icij instead of project eslint config

marieglr:
This PR removes the local config of eslint and replaces it with the one from: https://github.com/ICIJ/eslint-config-icij.

The purpose is to have homogenous code style accross vue.js projects.

Mains config changes are:
- removing .eslintrc file and eslint related plugins and packages
- adding an eslinConfig property in package.json simply extending the eslint-config-icij package

Main fixes in this PR are: 
- removing the file extensions in file imports, especially `.vue`and `.js`extensions
- reordering imports for external imports to be on top (libs..) and project imports below, separated by a line break
- one occurrence of nextTick used without a callback
- one occurence of a computed property used in a data property ---> did not manage to fix this one, you


src/components/NamedEntityInContext.vue
```diff
@@ -83,7 +83,8 @@ export default {
   },
   async mounted() {
     await this.loadContent()
-    await this.$nextTick(this.scrollToFirstMark)
+    await this.$nextTick()
+    await this.scrollToFirstMark()
```
caro3801, line 87: this.scrollToFirstMark is not async

src/components/filter/FilterBoilerplate.vue
```diff
@@ -229,6 +229,7 @@ export default {
   },
   data() {
     return {
+      // TODO: hasValues is a computed property, don't know how to fix this one
       collapseItems: this.collapsedIfNoValues && !this.hasValues,
```
marieglr, line 233: `hasValues` here is a computed property, I tried to move the whole `collapseItems` to ba a computer property but it breaks the implem. Do you have a better suggestion?

src/main.js
```diff
@@ -2,6 +2,8 @@ import { library as fortawesome } from '@fortawesome/fontawesome-svg-core'
 
 import * as icons from '@/utils/font-awesome'
 import { createCore } from '@/core'
+// TODO: add .scss in list of allowed extensions in eslint-config-icij
```
caro3801, line 5: Todo now or later?

src/components/filter/FilterBoilerplate.vue
caro3801: It seems that this is something coming from vue 1.x (https://v2.vuejs.org/v2/guide/migration.html#cache-false-deprecated). The migration guide recommands to transform it as a method and call it like a method
```
     L9: 'filter--has-values': hasValues()
     L233: collapseItems: this.collapsedIfNoValues && !this.hasValues(),
```

from:
```
computed:{
     hasValues: {
        cache: false,
        get() {
          return this.$store.getters['search/hasFilterValues'](this.filter.name)
        }
    }
}
```
to
```
methods:{
     hasValues(): {
          return this.$store.getters['search/hasFilterValues'](this.filter.name)
    }
}
```

caro3801 (changes requested):

src/main.js
marieglr: I will do it following this yes, but it needs for a new version of the package to be released. 

Do you think this single comment is a reason to block the merge of this PR with 'request changes' ? NB: you can stack your comments in a pull request review instead of making a new review for each comment. You can also comment without approving nor request changes if you don't want to approve but don't need for the next round of review to necessarily be made by you.

package.json
```diff
@@ -71,27 +71,17 @@
...
     "@vue/cli-plugin-unit-jest": "^4.5.13",
     "@vue/cli-service": "^4.5.13",
-    "@vue/eslint-config-standard": "^6.0.0",
     "@vue/test-utils": "^1.2.1",
     "babel-core": "7.0.0-bridge.0",
-    "babel-eslint": "^10.0.3",
     "babel-jest": "^26.6.3",
     "babel-plugin-transform-require-context": "^0.1.1",
     "canvas": "^2.9.3",
-    "eslint": "^7.30.0",
```
caro3801, line 87: this a test

src/components/filter/FilterBoilerplate.vue
```diff
@@ -519,7 +517,7 @@ export default {
       return get($infiniteLoadingState, method, noop)()
     },
     toggleItems() {
-      this.collapseItems = !this.collapseItems
```
caro3801, line 522: big problem delete please

caro3801 (approved):

PR #255, opened 2023-03-16
feat: adds alert informing tesseract is missing

mvanzalu:
Related to [this issue](https://github.com/ICIJ/datashare/issues/1013)
This PR adds an alert when OCR is check in indexation form.

Possible improvement : disable/grey out check button

### DIFF
--- a/src/components/ExtractingFormOcrControl.vue
+++ b/src/components/ExtractingFormOcrControl.vue
@@ -30,11 +30,14 @@ export default {
       }
       return 'default'
     },
+    isLanguageAvailable() {
+      return !!find(this.ocrLanguages, (language) => language.iso6392 === this.isoLang)
+    },
     isOcrLanguage() {
-      return (
-        (this.hasTesseract && !this.isoLang) ||
-        !!find(this.ocrLanguages, (language) => language.iso6392 === this.isoLang)
-      )
+      return !this.isoLang || this.isLanguageAvailable
+    },
+    shouldDisplayLanguageMessage() {
+      return this.hasTesseract && !this.isOcrLanguage
     },
     waitIdentifier() {
       return uniqueId('extracting-form-ocr-control-')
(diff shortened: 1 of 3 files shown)


src/components/ExtractingFormOcrControl.vue
```diff
@@ -18,7 +18,8 @@ export default {
   data() {
     return {
       textLanguages: [],
-      ocrLanguages: []
+      ocrLanguages: [],
+      hasTesseract: true
```
marieglr, line 22: This will always be true, no ?

src/components/ExtractingFormOcrControl.vue
```diff
@@ -30,7 +31,10 @@ export default {
       return 'default'
     },
     isOcrLanguage() {
-      return !this.isoLang || !!find(this.ocrLanguages, (language) => language.iso6392 === this.isoLang)
+      return (
+        (this.hasTesseract && !this.isoLang) ||
+        !!find(this.ocrLanguages, (language) => language.iso6392 === this.isoLang)
```
pirhoo, line 36: `!!find(this.ocrLanguages, (language) => language.iso6392 === this.isoLang` should be in a computed property

src/components/ExtractingFormOcrControl.vue
```diff
@@ -56,10 +60,14 @@ export default {
         this.textLanguages = textLanguages
         this.ocrLanguages = ocrLanguages
       } catch (e) {
-        this.$root.$bvToast.toast(this.$t('extractingLanguageFormControl.failedToRetrieveLanguages'), {
-          noCloseButton: true,
-          variant: 'danger'
-        })
+        if (e.response.status === 503) {
+          this.$set(this, 'hasTesseract', false)
```
pirhoo, line 64: for consistency I would have used:

```
this.hasTesseract = false
```

Or even better:

```
this.hasTesseract = e.response.status !== 503

if (this.hasTesseract) {
  this.$root.$bvToast.toast(this.$t('extractingLanguageFormControl.failedToRetrieveLanguages'), {
    noCloseButton: true,
    variant: 'danger'
  })
}
```

pirhoo (approved):
Great job! I like the fact that you used a 503 code!

I made some non-blocking suggestions.

src/components/ExtractingFormOcrControl.vue
pirhoo: Nop, hasTesseract is set to false when getting the method who gets the languages fails :)

src/components/ExtractingFormOcrControl.vue
mvanzalu: It works better yes

mvanzalu:
If we want to prevent the user from launching a OCR by disabling the switch button, what's the best thing to do ? Emit the 503 to the parent `ExtractingForm.vue` or rework both components ?

mvanzalu:
I think in `ExtractingForm` it's fine, I'll do another issue for this feature

PR #260, opened 2023-04-05
fix: path filter links color

pirhoo:
For some reason the class indicating "dark mode" on a filter was removed. This PR restore it and fix the class of tree view link with a deep selector.


caro3801 (approved):
Nice catch!

PR #261, opened 2023-04-05
fix: move jump to active local search on localSearchTerm

pirhoo:
This PR just move a bit of code that was conflicting with the page watch, and local search term watcher.


src/components/DocumentContent.vue
```diff
@@ -69,21 +69,16 @@ export default {
       this.localSearchTerm = value
     },
     localSearchTerm: throttle(async function (value) {
+      this.hasStickyToolbox = true
```
caro3801, line 72: Make sense indeed!

caro3801 (approved):

PR #264, opened 2023-05-12
fix: ensure changing target language is reactive

pirhoo:
This fix this bug: https://github.com/ICIJ/datashare/issues/1063

The watcher on `targetLanguage` was "recooking" extract text because we used to load content in the recooking method. Since we moved the loading to a higher level method (`activateContentSlice`) we needed to call this new method instead.


caro3801 (approved):
Perfect

PR #278, opened 2023-06-30
feat: merged download buttons

pirhoo:
This PR resolves this https://github.com/ICIJ/datashare/issues/1100.

### Before

Screenshot 2023-06-30 at 12-47-35 Document - Datashare

### After

Screenshot from 2023-06-30 13-15-48


marieglr (approved):
Nice :+1: 

I would have the "download" icon for all kind of downloads, I find it more clear to understand in a glimps, and the specific meaning of the other icons to subtle to understand in this context, but that's just an opinion and a minor one. 

(Also you should coonfigure the linter in your IDE :speak_no_evil: )

pirhoo:
Thanks! I updated the icons :)

PR #286, opened 2023-07-17
feat: batch search form page

pirhoo:
## PR description

In order to anticipate on future enhancements, this PR aims at moving the Batch Search Form to a separated page (instead of a modal).

Screenshot 2023-07-17 at 19-08-22 Start a new batch search - Datashare

## Changes

* Create a new page `NewBatchSearch` to hold the form
* Create a new page `BatchSearch` to hold all batch search related views 
* Rename `BatchSearch` to `BatchSearchList`
* Make `NewBatchSearch`, `BatchSearchList` and `BatchSearchResult` children of the same route
* Remove `BatchSearchForm` from the modal

## Limitations

* The `batch-search` route should never be navigated directly (it's empty), hence why I edited all links to use `batch-search-list` ;
* `NewBatchSearch` page name might be confusing although its syntax is more correct than `BatchSearchNew`.

### DIFF
--- a/src/pages/NewBatchSearch.vue
+++ b/src/pages/NewBatchSearch.vue
@@ -26,10 +26,7 @@ export default {
         <fa icon="angle-left" class="mr-1" fixed-width />
         {{ $t('batchSearch.title') }}
       </router-link>
-      <span class="border-left border-dark pl-3 ml-3 font-weight-bold">
-        {{ $t('newBatchSearch.title') }}
-      </span>
     </div>
-    <batch-search-form hide-title @submit="submit"></batch-search-form>
+    <batch-search-form @submit="submit"></batch-search-form>
   </div>
 </template>
--- a/src/pages/NewBatchSearch.vue
+++ b/src/pages/NewBatchSearch.vue
@@ -1,0 +1,35 @@
+<script>
+import BatchSearchForm from '@/components/BatchSearchForm'
+
+export default {
+  name: 'NewBatchSearch',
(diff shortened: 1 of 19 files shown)


src/pages/Tasks.vue
```diff
@@ -75,14 +70,17 @@ export default {
       },
       set(value) {
         const name = this.tabRoutes[value]
+        const resolvedPath = this.$router.resolve({ name })?.route?.path
+        // Must be the current route or one of its children
+        const isMatchedRoute = this.$route.matched.some((m) => m.path.startsWith(resolvedPath))
```
caro3801, line 75: Smart :sunglasses:

src/pages/NewBatchSearch.vue
```diff
@@ -0,0 +1,35 @@
...
+</script>
+
+<template>
+  <div class="container py-4">
+    <div class="mx-1 mb-2 d-flex">
+      <router-link :to="batchSearchRoute">
+        <fa icon="angle-left" class="mr-1" fixed-width />
+        {{ $t('batchSearch.title') }}
+      </router-link>
+      <span class="border-left border-dark pl-3 ml-3 font-weight-bold">
```
caro3801, line 29: As Soline suggested it could be interesting to move the title below the "go back" button

caro3801 (approved):
Nice addition thanks
We might need to adapt BatchSearchResults go back button to the new batch search list route name

PR #298, opened 2023-08-01
feat: allow widgets settings to use a factory function or a class

pirhoo:
## PR description

This PR aims at letting plugin define their own widgets (https://github.com/ICIJ/datashare/issues/1148).

## API

The new API can be used as follow (using the same `registerWidget` method):

```js
core.registerWidget({
  name: 'test',
  type(WidgetEmpty) {
    return class WidgetTest extends WidgetEmpty {
      get component() {
        return {
          name: 'WidgetTest',
          template: '<div>WidgetTest</div>'
        }
      }
    }
  }
})
```


caro3801 (approved):
LGTM

PR #307, opened 2023-08-05
feat: sort document search by file name

mvanzalu:
**Goal**
This PR is related to [this issue](https://github.com/ICIJ/datashare/issues/1103)


mvanzalu:
So far the current implementation allows the user to filter searches by document name. However it sorts in the following order : Numbers and special chars - uppercase letter - lowercase letter. We tried another ES settings to allow to sort in a natural way (case insensitive) but it breaks the search functionality.

- [ ] Sort by name
- [ ] Case insensitive sort

pirhoo:
This seems good to me but I would like to test the performances first.

mvanzalu:
Sorting is working in specs tests, however I get strange results in manual local test

PR #308, opened 2023-08-07
refactor : extract axios mock and Api instance from the tests

caro3801:
Refactor to mock only API methods instead of mocking axios.
Core is now initialized with a null method to enforce the mocking of the API methods used in the test and only them.
This allows to be less impacted by axios implementation (which is fully tested in api/index.js) and more focused on how the api methods are called.

### DIFF
--- a/tests/unit/specs/components/TreeView.spec.js
+++ b/tests/unit/specs/components/TreeView.spec.js
@@ -36,8 +36,8 @@ const HOME_TREE_WIN = {
 describe('TreeView.vue', () => {
   describe('Posix', () => {
     const { index, es } = esConnectionHelper.build()
-
-    const { config, i18n, localVue, store, wait } = Core.init(createLocalVue()).useAll()
+    const api = { tree: jest.fn() }
+    const { config, i18n, localVue, store, wait } = Core.init(createLocalVue(), api).useAll()
     const propsData = {
       projects: [index],
       path: '/home/foo',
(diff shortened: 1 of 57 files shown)


tests/unit/specs/components/TreeView.spec.js
```diff
@@ -93,7 +93,7 @@ describe('TreeView.vue', () => {
     })
 
     it('should be a display a correct basename', async () => {
-      wrapper.vm.$core.api.tree = jest.fn().mockResolvedValue(HOME_TREE)
+      // api.tree.mockResolvedValue(HOME_TREE)
```
pirhoo, line 96: I guess we can delete this :wastebasket:

tests/unit/specs/core/Core.spec.js
```diff
@@ -5,133 +5,126 @@ import VueI18n from 'vue-i18n'
...
-
-  it("should return first user project when user doesn't have the default project", () => {
-    const core = Core.init(localVue).useAll()
-    core.config.set('projects', [{ name: 'user_project' }])
-    core.config.set('defaultProject', 'default_project')
-    expect(core.getDefaultProject()).toEqual('user_project')
+    expect(core.i18n).not.toBeInstanceOf(VueI18n)
+  })
+
+  describe('Call useAll on Core', () => {
```
pirhoo, line 44: Good call

pirhoo (approved):

caro3801:
Thanks for the review

PR #315, opened 2023-08-30
feat: new project page

caro3801:
This PR resolves: ICIJ/datashare#1160 and resolves  ICIJ/datashare#1159
It adds in local mode:
- a button on the project page to create a new project 
- a tab to launch index task AKA "add documents" per project
- a tab to launch named entities extraction AKA "find named entities" per project

This involves the refactoring of the tasks forms to made display them as pages instead of modal with props change:
- dark background option
- project name option to directly select the concerned project
- add of a submit event to handle toast notifications when the task is launched

Fix: 
In `FindNamedEntitiesForm.vue` the radio button was previously not selected due to lowercase/uppercase mismatch of the string CORENLP.

Screenshots:
Screenshot from 2023-08-30 15-26-45

Screenshot from 2023-08-30 15-26-59

Screenshot from 2023-08-30 15-27-49

Screenshot from 2023-08-30 15-27-56

### DIFF
--- a/src/components/FindNamedEntitiesForm.vue
+++ b/src/components/FindNamedEntitiesForm.vue
@@ -159,8 +159,8 @@ export default {
         this.$wait.start('launch ner task')
         await this.$store.dispatch('indexing/submitFindNamedEntities')
         this.$emit('submit', { error: false })
-      } catch (e) {
-        this.$emit('submit', { error: e })
+      } catch (error) {
+        this.$emit('submit', { error })
       } finally {
         this.$wait.end('launch ner task')
         this.$store.commit('indexing/resetFindNamedEntitiesForm')
--- a/src/components/FindNamedEntitiesForm.vue
+++ b/src/components/FindNamedEntitiesForm.vue
@@ -135,6 +135,7 @@ export default {
         await this.$store.dispatch('indexing/submitFindNamedEntities')
       } finally {
         this.$store.commit('indexing/resetFindNamedEntitiesForm')
+        this.disabled = false
         this.$emit('submit')
       }
     },
--- a/src/components/FindNamedEntitiesForm.vue
+++ b/src/components/FindNamedEntitiesForm.vue
@@ -54,6 +54,7 @@
       </div>
       <div class="find-named-entities-form__footer mt-4 row no-gutters">
         <slot name="footer" :disabled="disabled">
+          disabled {{ disabled }}
           <div class="col text-right">
             <b-btn variant="primary" class="font-weight-bold" type="submit" :disabled="disabled">
               {{ $t('indexing.go') }}
@@ -113,7 +114,7 @@ export default {
   },
   computed: {
     disabled() {
-      return this.pipeline && !this.$wait.waiting('load ner pipelines')
+      return !this.pipeline && (!this.$wait.waiting('load ner pipelines') || !this.$wait.waiting('launch ner task'))
     },
     offline: {
       set(value) {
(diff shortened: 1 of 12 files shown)


src/components/FindNamedEntitiesForm.vue
```diff
@@ -78,20 +86,36 @@ export default {
   mixins: [utils],
   props: {
     /**
-     * Callback function to call when the form have been submitted (this should be replaced by an event in future versions).
+     * Dark mode background option
+     */
+    dark: {
+      type: Boolean
```
bamthomas, line 92: is dark mode a new feature in ds-client?

bamthomas (approved):

src/components/FindNamedEntitiesForm.vue
```diff
@@ -78,20 +86,36 @@ export default {
...
-      type: Function,
-      default: noop
+    projectName: {
+      type: String,
+      default: null
+    },
+    /**
+     * Disable project selection select input
+     */
+    disableProjectSelection: {
```
pirhoo, line 104: It's good you block user from changing the project but I would completely hide the project input instead. No need to show it as it's obvious which project is being selected.

src/components/FindNamedEntitiesForm.vue
```diff
@@ -78,20 +86,36 @@ export default {
...
+      type: Boolean,
+      default: false
     }
   },
   data() {
     return {
       pipelines: [],
-      disabled: false
+      waiting: false
```
pirhoo, line 112: As far as I can see, `this.waiting` is always false. Maybe you meant to use the `this.$wait.start('load ner pipelines')`?

src/components/FindNamedEntitiesForm.vue
```diff
@@ -105,35 +129,39 @@ export default {
...
     showProjectSelector() {
       return this.$core.projects.length > 1 || this.defaultProject !== this.$config.get('defaultProject')
     }
   },
   async mounted() {
-    this.$wait.start('load ner pipelines')
+    if (this.$core.findProject(this.projectName)) {
+      this.defaultProject = this.projectName
+    }
+    this.$wait.start('load ner pipelinesccccc')
```
pirhoo, line 151: Typo here with `load ner pipelinesccccc`

src/components/FindNamedEntitiesForm.vue
```diff
@@ -105,35 +129,39 @@ export default {
...
     this.$set(this, 'pipelines', this.handlePipelinesTranslation(values(pipelines).map(lowerCase)))
     this.$wait.end('load ner pipelines')
   },
   methods: {
     async submitFindNamedEntities() {
-      this.disabled = true
       try {
         await this.$store.dispatch('indexing/submitFindNamedEntities')
+        this.$emit('submit', { error: false })
+      } catch (e) {
```
pirhoo, line 161: If you name the error parameter `error` you can use it directly in the next instruction: `this.$emit('submit', { error })`

pirhoo (changes requested):
I see minor problems, you should be able to solve them very quickly :)

src/components/FindNamedEntitiesForm.vue
caro3801: Yes  good catch thanks

src/components/FindNamedEntitiesForm.vue
caro3801: ok I'm changing it to hide the input

caro3801:
Hiding project selector instead of disabling it as suggested by @pirhoo 
Screenshot from 2023-08-30 17-28-53

PR #318, opened 2023-09-01
feat: adds Title and Title Norm in Document Tab Details

mvanzalu:
Related to [this issue](https://github.com/ICIJ/datashare/issues/1173)

### DIFF
--- a/src/components/document/DocumentTabDetails.vue
+++ b/src/components/document/DocumentTabDetails.vue
@@ -175,10 +175,16 @@ export default {
           }
         },
         {
-          name: 'metadata.tika_metadata_resourcename',
-          label: this.$t('document.name'),
-          trClass: 'document__content__basename',
-          value: this.document.meta('resourcename')
+          name: 'title',
+          label: this.$t('document.title'),
+          trClass: 'document__content__title',
+          value: this.document.title
+        },
+        {
+          name: 'title',
+          label: this.$t('document.titleNorm'),
+          trClass: 'document__content__title-norm',
+          value: this.document.titleNorm
         },
         {
           name: 'path',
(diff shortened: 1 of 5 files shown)


src/components/document/DocumentTabDetails.vue
```diff
@@ -175,10 +175,16 @@ export default {
...
-          name: 'metadata.tika_metadata_resourcename',
-          label: this.$t('document.name'),
-          trClass: 'document__content__basename',
-          value: this.document.meta('resourcename')
+          name: 'title',
+          label: this.$t('document.title'),
+          trClass: 'document__content__title',
+          value: this.document.title
+        },
+        {
```
pirhoo, line 183: I don't think we need to display the `titleNorm`, it is only for sorting.

src/lang/en.json
```diff
@@ -400,6 +400,8 @@
     "title": "Document",
     "document": "Document",
     "name": "Document name",
+    "title": "Title",
+    "titleNorm": "Title Normalized",
```
pirhoo, line 404: If we don't display it in the detail tab, we don't need this.

tests/unit/specs/components/document/DocumentTabDetails.spec.js
```diff
@@ -189,4 +189,19 @@ describe('DocumentTabDetails.vue', () => {
...
+    wrapper = mount(DocumentTabDetails, {
+      i18n,
+      localVue,
+      store,
+      router,
+      propsData: { document: store.state.document.doc }
+    })
+
+    expect(wrapper.find('.document__content__title').exists()).toBeTruthy()
+    expect(wrapper.find('.document__content__title-norm').exists()).toBeTruthy()
```
pirhoo, line 205: Same here

pirhoo (changes requested):

src/components/document/DocumentTabDetails.vue
mvanzalu: Indeed !

pirhoo (approved):

PR #322, opened 2023-10-05
feat: use field endpoint for each field

pirhoo:
This avoids loading the entire mapping for all user's projects (see https://github.com/ICIJ/datashare/issues/1204).


caro3801 (approved):
Very much appreciated feature! Thanks

PR #326, opened 2023-10-09
fix: batch search query route

pirhoo:
## PR description

This ensure we can click on a query from a batch search to be redirected to the general search page.

## Changes

* fix: use the correct name in the route property ("indices" instead of "index")
* refactor: use the "indices" property so it's always a string


bamthomas (approved):

caro3801 (changes requested):
nice refactor, it improves control over indices formatting
It would be nice to add either default value or required attribute to props.
The duplicate navigation error should be checked too.

PR #332, opened 2023-10-25
feat: add widget recommended by

pirhoo:
## PR description

This PR adds a new "recommended by" widget to the project page (see https://github.com/ICIJ/datashare/issues/1228).

## Preview 

Large screen:

Screenshot 2023-10-25 at 11-45-15 Datashare

Small screen:

Screenshot 2023-10-25 at 11-56-09 Datashare


src/components/DocumentSlicedName.vue
```diff
@@ -98,7 +106,7 @@ export default {
       return { id: this.document.source.rootDocument || this.document.id }
     },
     baseComponent() {
-      return this.hasActiveTextTruncate ? ActiveTextTruncate : 'span'
+      return this.hasActiveTextTruncate() ? ActiveTextTruncate : 'span'
```
caro3801, line 109: `hasActiveTextTruncate` looks like a computed property to me

src/components/widget/WidgetRecommendedBy.vue
```diff
@@ -0,0 +1,202 @@
+<template>
+  <div class="widget widget--recommended-by">
+    <div class="widget__header d-flex align-items-center" :class="{ 'card-body': widget.card }">
+      <fa icon="users" fixed-width class="mr-2" />
+      <h3 class="m-0 p-0 h5">Latest recommended documents</h3>
```
caro3801, line 5: I would add an i18n key for translation to "Latest recommended documents"

src/components/widget/WidgetRecommendedBy.vue
```diff
@@ -0,0 +1,202 @@
...
+          <div class="flex-grow-1">
+            <div class="widget__list__item__label d-flex align-items-start">
+              <document-sliced-name wrap :document="document" class="text-primary" />
+              <div class="ml-auto pl-1 d-flex align-items-center text-muted flex-shrink-0">
+                <user-display :username="user.id" hide-link class="text-nowrap" />
+                <span
+                  v-if="creationDate"
+                  v-b-tooltip
+                  class="ml-1 text-nowrap"
+                  :title="creationDate | humanLongDate($i18n.locale)"
```
caro3801, line 30: Since filters will be deprecated in Vue 3 I would suggest to use computed properties instead (but I also like this feature so I don't mind if stay as is).

src/main.scss
```diff
@@ -99,6 +99,23 @@ body {
   }
 }
 
+// Truncated element in a flexbox child must not fully collapse when the
+// flex container is resized down. 
+// @see https: //drafts.csswg.org/css-flexbox/#min-size-auto
```
caro3801, line 104:
```suggestion
// @see https://drafts.csswg.org/css-flexbox/#min-size-auto
```
what was the behavior before?

caro3801 (approved):
Nice feature, good looking widget and great code, what else !

src/components/DocumentSlicedName.vue
pirhoo: It's not, hence the confusion before! We should probably refactor this. This change here is just a bug fix.

src/components/widget/WidgetRecommendedBy.vue
pirhoo: Ah yes, you're right I forgot.

src/components/widget/WidgetRecommendedBy.vue
pirhoo: We still use the filter in many places so I guess we can refactor them all at once!

PR #338, opened 2023-12-04
refactor: change TaskView model name -> id + name 

bamthomas:
This is to adapt BatchDownload interface to refactor made with  ICIJ/datashare#1081

### DIFF
--- a/src/components/BatchDownloadActions.vue
+++ b/src/components/BatchDownloadActions.vue
@@ -7,6 +7,12 @@ export default {
     /**
      * Name of the batch download's task
      */
+    id: {
+      type: String
+    },
+    /**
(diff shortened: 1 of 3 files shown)


src/components/BatchDownloadActions.vue
```diff
@@ -4,6 +4,12 @@ import { uniqueId } from 'lodash'
 export default {
   name: 'BatchDownloadActions',
   props: {
+    /**
+     * Name of the batch download's task
+     */
+    id: {
+      type: String
```
caro3801, line 11:
```suggestion
      type: String,
      required: true
```

caro3801 (approved):

PR #340, opened 2023-12-11
feat: adds tags in batch search

mvanzalu:
PR related to [this issue](https://github.com/ICIJ/datashare/issues/1071) and [this one](https://github.com/ICIJ/datashare/issues/1248).

The PR adds an input to select tags in the batch search form and a switch button to include/exclude tags for the search.


tests/unit/specs/components/BatchSearchForm.spec.js
```diff
@@ -236,6 +263,102 @@ describe('BatchSearchForm.vue', () => {
     })
   })
 
+  describe('Tags suggestions', () => {
+    it('should display suggestions', () => {
+      expect(wrapper.find('.batch-search-form__tags__suggestions').exists()).toBe(true)
+    })
+
+    it('should hide suggestions', () => {
+      wrapper.setData({ suggestionTags: ['suggestion_01', 'suggestion_02', 'suggestion_03'] })
```
caro3801, line 272: setData is async : every setData should be preceded with await. This can prevent async side effects when test are ran in parallel.

src/components/BatchSearchForm.vue
```diff
@@ -317,102 +370,152 @@ export default {
...
       })
-      this.$set(this, 'suggestionFileTypes', fileTypesWithoutSelection)
     }, 200),
     searchFileType() {
       if (this.selectedFileType) {
         this.fileTypes.push(this.selectedFileType)
         this.hideSuggestionsFileTypes()
-        this.$set(this, 'fileType', '')
+        this.fileType = initData.fileType
         if (this.$refs && this.$refs.fileType) this.$refs.fileType.focus()
```
caro3801, line 430:
```suggestion
        this.$refs?.fileType?.focus()
```

src/components/BatchSearchForm.vue
```diff
@@ -317,102 +370,152 @@ export default {
...
+      this.suggestionTags = filter(tags, (item) => {
+        return !includes(this.tags, item)
+      })
+    }, 200),
+    searchTag() {
+      if (this.selectedTag) {
+        this.tags.push(this.selectedTag)
+        this.hideSuggestionsTags()
+        this.tag = initData.tag
+        if (this.$refs && this.$refs.tag) this.$refs.tag.focus()
```
caro3801, line 487:
```suggestion
        this.$refs?.tag?.focus()
```

caro3801 (approved):
Great feature addition thanks!

tests/unit/specs/components/BatchSearchForm.spec.js
mvanzalu: Thanks, I'm correcting it

src/components/BatchSearchForm.vue
```diff
@@ -317,102 +370,156 @@ export default {
...
+    async showAdvancedFilters() {
+      await Promise.all([this.retrieveFileTypes(), this.retrieveTags()])
     }
   },
   created() {
-    this.$set(this, 'selectedProjects', this.defaultSelectedProjects)
+    this.selectedProjects = this.defaultSelectedProjects
   },
   methods: {
+    createQueryTemplate() {
```
caro3801, line 408: This can be transformed as a computed prop
We can keep it as a **method** with this signature:
```js
createQueryTemplate(filters){
  const textFilters = filters.map(f=>{
    const textFilter = new FilterText({ name: f.name, key: f.key, forceExclude: f.forceExclude })
    textFilter.values = f.values
    return textFilter
  })

  const { query } = this.$core.api.elasticsearch
    .rootSearch(textFilters, TEMPLATE_VALUE.QUERY)
    .build()
  return JSON.stringify(query)
}
```
and have a **computed** prop 
```js
queryTemplate(){
  const filters = [
    { name: 'tags', key: 'tags', forceExclude: this.forceExclude, values: this.tags},
    { name: 'contentType', key: 'contentType' , forceExclude: false, values: this.fileTypes },
    {name: 'path', key: 'byDirname', forceExclude: false, values: this.paths }
  ]

  return createQueryTemplate(filters)
}
```
This will be extensible if we want to add exclude filters on other attributes

caro3801 (changes requested):
great improvements. The requested changes will prevent repetition and allow extensible version of the feature

src/components/BatchSearchForm.vue
mvanzalu: Good way to refactor. The only catch is that `path`is a `FilterPath` type instead of `FilterText`

caro3801: Oups therefore it's possible to add switch case in the map  :

```js
...
const textFilters = filters.map(f=>{ 
  let filter;
  switch (f.name) {
   case 'path':
    filter = new FilterPath({ name: f.name, key: f.key, forceExclude: f.forceExclude }) 
    break;
  case 'contentType':
  case 'tags:
    filter = new FilterText({ name: f.name, key: f.key, forceExclude: f.forceExclude })
    break;
  default: 
    throw new Error("unsupported query filter")
  }
  filter.values = f.values 
  return filter
 })
...

caro3801 (approved):

Issue #341, opened 2023-12-15
[Front-end bugs] cannot log in normally when the mode is server mode.


a3sroot:
https://github.com/ICIJ/datashare/issues/1279

a3sroot:
BasicAuthAdaptorFilter 
/api/user/me 
There is no uid field.

PR #342, opened 2023-12-16
fix: basicAuthBug

a3sroot:
#341


a3sroot:
@Soliine

pirhoo:
Hello @a3sroot, thanks for submitting the PR but it's more a workaround rather than fixing the root issue (on the backend side). Our team will fix it, thanks for your patience.

PR #363, opened 2024-12-17
refactor: change api endpoint for batchsearch creation/copy 

bamthomas:
see https://github.com/ICIJ/datashare/issues/1623


caro3801 (commented):
Thanks! I agree with the changes from `/batch` to `/task` but I would suggest to use `/task/batchsearch` instead of `/task/search` . Wdyt?

PR #389, opened 2025-07-18
Chore/update murmur

caro3801:


pirhoo (changes requested):
Thanks for this work!

Murmur's plugins are installed globally so in theory you don't have to do `import { ButtonIcon } from '@icij/murmur-next'` every time you use it. I've made a suggestion to ensure retro compatibility with the plugins.

pirhoo (approved):

PR #396, opened 2025-08-18
fix: delete task confirmation

mvanzalu:
This PR aims to fix the delete task buttons in Document Task page, it additionally adds a confirmation before deleting.
The PR also implements the latter in the Entity Task page.

### DIFF
--- a/src/views/Task/TaskDocuments/TaskDocumentsList.vue
+++ b/src/views/Task/TaskDocuments/TaskDocumentsList.vue
@@ -31,10 +31,10 @@ const { t } = useI18n()
 async function stopTask(uuid) {
   return taskStore.stopTask(uuid)
 }
-const successMessage = t('task.remove.success')
-const errorMessage = t('task.remove.error')
+
 function remove(id) {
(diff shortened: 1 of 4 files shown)


pirhoo (changes requested):
Thanks @mvanzalu, I made some cosmetic suggestions to ensure style consistency but the PR itself is a very nice addition!

src/views/Task/TaskDocuments/TaskDocumentsList.vue
```diff
@@ -45,6 +45,12 @@ function isRunning(item) {
   return item.state === TASK_STATUS.RUNNING
 }
 
+function removeWithConfirmation(id, callback) {
+  afterConfirmation(async () => {
```
pirhoo, line 49: Because `afterConfirmation` is async, it's important to have a `return` here or the promise will create an "unhandled promise rejection" warning.

```suggestion
  return afterConfirmation(async () => {
```

pirhoo (approved):
All good, thanks the changes!

PR #410, opened 2025-11-13
fix: mounted data dir

pirhoo:
As described in icij/datashare#1923, this PR fixes a bad usage of the "mounted data dir" option which was overriding the actual value of some path in various places (project form, project details widget, document metadata tab).


caro3801 (approved):
Thanks, for the changes, looks good to have factorize the path on one place.
Maybe adding tests on the composable useDataDir's function `getMountedPath`could add some resilience with path formatting errors (or windows format) 
side question: WidgetProject has been removed because it was not used?

PR #419, opened 2026-01-22
feat: admin update project metadata

caro3801:
Related to feature ICIJ/datashare#1919

:warning: Merge first backend changes https://github.com/ICIJ/datashare/pull/2011  

Changes: in server mode, admins are now allowed to update project metadata in project edit page

The project view when user is not admin:
image
User still can't access edit page http://localhost:9009/#/projects/local-datashare/edit (Something went wrong)

when user is admin, they see the Edit button:
image

They can access the edit project page:
image

Project is updated when hitting save (project must exist in the DB first)
image

### DIFF
--- a/src/components/Display/DisplayRole.vue
+++ b/src/components/Display/DisplayRole.vue
@@ -23,7 +23,7 @@ const role = computed(() => formatRole(props.value))
 </script>
 
 <template>
-  <span class="d-inline-flex gap-1">
+  <span class="display-role d-inline-flex gap-1">
     <app-icon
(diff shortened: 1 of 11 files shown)


pirhoo (changes requested):
That's so cool to see it live!

I've made a few improvement suggestions. Also, I wonder if it wouldn't be simpler for the user to only see the highest rank role. For instance display only "editor" if they are "viewer" and "editor" or "admin" if they are "admin", "editor" and "viewer".

PR #428, opened 2026-03-26
 fix: correct "All" entry count in filter panels

caro3801:
**fix: correct "All" entry count in filter panels**
related to https://github.com/ICIJ/datashare/issues/2019 
The "All" entry in filter panels (content type, language, recommended by, starred, etc.) was incorrectly showing `searchStore.total` - the global document count - instead of a count meaningful to each filter type.

**Changes:**

- **`FilterType.vue`**: Computes `bucketsTotal` as the sum of `doc_count` across all loaded aggregation buckets and passes it as `:count` to `FilterTypeAll`. Returns `null` before any buckets are fetched (keeps count hidden until data arrives). 

- **`FilterTypeAll.vue`**: Replaces the `computedTotal`/`searchStore` coupling with a `count` prop (Number, default `null`). The `hadTotal` guard now uses an immediate watcher on `count` to prevent flickering on first render. Each filter type controls its own count semantics.

- **`FilterTypeRecommendedBy.vue`**: Overrides the `#all` slot to pass `recommendedStore.total` (sum of all recommendation counts across selected indices) as the "All" count. Uses a `waitFor` so count is `null` until the fetch resolves, preventing a premature `0`.

- **`FilterTypeStarred.vue`**: Overrides the `#all` slot to pass the real document total from `getTotal()` as the "All" count. Uses `waitFor` for flicker-prevention reason. Also fixes a copy-paste bug where the "Not starred" entry had `name="starred"`. Exposes `fetch` for testability.

---
AI disclaimer
Used to write the description more cleanly
use agent to do first draft of unit test (TDD like)

### DIFF
--- a/src/components/Filter/FilterType/FilterTypeAll.vue
+++ b/src/components/Filter/FilterType/FilterTypeAll.vue
@@ -1,5 +1,5 @@
 <script setup>
-import { computed, ref, toRef, watch } from 'vue'
+import { toRef } from 'vue'
 import { useI18n } from 'vue-i18n'
 
 import FiltersPanelSectionFilterEntry from '@/components/FiltersPanel/FiltersPanelSectionFilterEntry'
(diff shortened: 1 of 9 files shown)


pirhoo (changes requested):
Thanks for this! Unfortunately it doesn't work as expected (I tested it manually to confirm my analysis of the code) and the count is now wrong not only for the entities filter, but also for the other filters if they have unloaded buckets pages:

Image

You should focus on overriding the count display only on named entities filters (probably by creating a new FilterTypeEntities component). The test and calculation you made works because you are only testing it with component that actually do aggregation over `type:Document` and because all their bucket pages are loaded. The tricky part of the entities filter is the fact that the aggregation is done on `type:NamedEntites`.

pirhoo (approved):

PR #433, opened 2026-04-09
feat: search results performances

pirhoo:
To address performances issues raised in https://github.com/ICIJ/datashare/issues/2109. This PR tries to reduce HTTP request during search by using 3 strategies:

1. **Lazy download status fetching**: Download status and translations are no longer fetched for every document on mount. Instead, each entry only fetches when it scrolls into view, and the popover only fetches when opened for the first time.
2. **Search deduplication**: Search results were queried twice on every page load due to two route watchers firing for the same navigation. Concurrent calls now share a single request.
3. **Document notes deduplication**: Concurrent calls to fetch notes for the same project now share a single API request instead of each triggering its own.

## Changes

- feat: add `immediate` option to `useDocumentDownload` composable, allowing callers to opt out of automatic `watchEffect`-driven status fetching on mount
- feat: expose a `fetchStatuses` function from `useDocumentDownload` that runs both download status and translation status checks in parallel
- perf: defer status fetching in `DocumentActionsGroupEntryDownload` until the element becomes visible, using `useElementVisibilityOnce`
- perf: defer status fetching in `DocumentDownloadPopover` until the popover is activated (opened for the first time) rather than on mount
- fix: deduplicate concurrent `refresh()` calls in the search store by sharing an in-flight promise, preventing duplicate `searchDocs` API calls when multiple triggers fire for the same navigation
- fix: deduplicate concurrent `fetchNotesOnce()` calls in the document notes store with a per-project promise cache, preventing redundant `retrieveNotes` API calls
- test: add tests covering deduplication of concurrent calls in `useDocumentDownload`, the search store, and the document notes store

## AI Disclosure

Only used to write/format the PR.

### DIFF
--- a/src/components/Document/DocumentActionsGroup/DocumentActionsGroupEntryDownload.vue
+++ b/src/components/Document/DocumentActionsGroup/DocumentActionsGroupEntryDownload.vue
@@ -1,5 +1,6 @@
 <script setup>
-import { computed, nextTick } from 'vue'
+import { computed, nextTick, useTemplateRef } from 'vue'
+import { whenever } from '@vueuse/core'
 import { useI18n } from 'vue-i18n'
 
 import IPhDownloadSimple from '~icons/ph/download-simple'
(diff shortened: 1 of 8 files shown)


caro3801 (approved):
Great request optimizations thanks!
On a small sample (22 docs), number of XHR queries are already divided  by 5!

PR #437, opened 2026-04-29
feat: add option to select search operator

caro3801:
Implements ICIJ/datashare#1497 - adds an option to switch Elasticsearch's default_operator between OR (default) and AND.                                                     
                                                                                                                                                                               
  - New setting: a searchOperator setting (`OR` , `AND`) added to SearchSettings, stored in `appStore` under `views.search`, with `SETTINGS_VERSION` bumped to 1 to trigger a clean hydration on upgrade.                                                                                                                                                        
  - Elasticsearch integration: the operator is threaded from the `app store` -> `searchDocs` -> `_buildSearchBody` -> `_applyQueryString`, where it sets `default_operator` in the `query_string` DSL clause. Batch download and download size estimation also respect the operator.                                                                              
  - Route persistence: `searchOperator` is included in `toRouteQuery` so it appears in the URL and is restored when navigating back or sharing a link. Invalid values in the URL are silently ignored and fall back to `OR`.
  - Reactivity: a `watchOperator` watcher in `useSearchFilter` triggers `refreshRouteFromStart` whenever the operator changes, so results update immediately after toggling.
  - UI display: `SearchParameterQueryAst` now resolves `'<implicit>'` AST operators to the active store value, so the query breadcrumb always shows the effective operator (OR or  AND). 


<img width="1517" height="487" alt="image" src="https://github.com/user-attachments/assets/4c4d54cb-c974-45aa-9508-96fd523c08bd" />

<img width="1517" height="487" alt="image" src="https://github.com/user-attachments/assets/65b927d8-302e-455a-9680-0994268a6d84" />

### DIFF
--- a/src/store/modules/app.js
+++ b/src/store/modules/app.js
@@ -10,7 +10,7 @@ import { SEARCH_OPERATORS } from '@/enums/searchOperators'
  * will require resetting user settings to defaults. This is useful if for instance you add new settings
  * or change the structure of existing settings and want to ensure users get the latest defaults.
  */
-const SETTINGS_VERSION = 0
+const SETTINGS_VERSION = 1
 
 /**
  * Defines the application-wide store for managing UI state and user preferences.
--- a/src/store/modules/app.js
+++ b/src/store/modules/app.js
@@ -3,6 +3,7 @@ import { ref, reactive } from 'vue'
 import { defineStore } from 'pinia'
 
 import { LAYOUTS } from '@/enums/layouts'
+import { SEARCH_OPERATORS } from '@/enums/searchOperators'
 
 /**
  * Version of the settings schema. Increment this value when making changes to the settings structure which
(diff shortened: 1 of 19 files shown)


src/store/modules/app.js
```diff
@@ -3,13 +3,14 @@ import { ref, reactive } from 'vue'
...
 import { LAYOUTS } from '@/enums/layouts'
+import { SEARCH_OPERATORS } from '@/enums/searchOperators'
 
 /**
  * Version of the settings schema. Increment this value when making changes to the settings structure which
  * will require resetting user settings to defaults. This is useful if for instance you add new settings
  * or change the structure of existing settings and want to ensure users get the latest defaults.
  */
-const SETTINGS_VERSION = 0
+const SETTINGS_VERSION = 1
```
mvanzalu, line 13: Question: i don't really understand the use of this constant. I assume it's recent

pirhoo: This is to force the settings cache to refresh when there is a new setting.

mvanzalu (approved):

pirhoo (approved):
Simple and efficient, well done!

PR #443, opened 2026-05-11
feat: add project users management tab (create, roles, delete)

caro3801:
:warning:  Merge #ICIJ/datashare#2259 before

## Feature: Project Users tab
Requires #ICIJ/datashare#2259 to test the feature

New `project.view.edit.users` route with a sortable, paginated, searchable user list, server-side search/sort/pagination, inline + batch role editing (with admin-promotion confirmation), row actions (copy/delete, self-delete disabled), and a create-user modal gated to form/basic auth.

**Intent:** self-service user/role management inside a project.

### "Behavior by auth mode / backend" 

- **oauth**: `isAuthWithUsersProvider` is false → no create/password UI, delete always uses `revokeUserRole` (role removal only, never account deletion).
- **form/basic + database-backed repository**: full create/delete via `createUser`/`deleteUser`.
- **form/basic + redis-backed (or non-durable) repository**: same UI shown (gating is only on auth mode, not repository type), but write support isn't guaranteed — flagged as worth confirming server-side, since the frontend can't detect this and just surfaces whatever error comes back.

## Refactor: User & role API surface (breaking)
- Removed `getProjectPolicies`/`saveProjectPolicy`/`removeProjectPolicy`; added `getUsers`/`grantUserRole`/`revokeUserRole`/`createUser`/`deleteUser` against new `/api/users` endpoints.
- Added `ROLE_LOWERCASE` — role values sent/received by these APIs are now lowercase, not the existing uppercase `ROLE` enum.
- `useAuth().isUsersProvider` renamed to `isAuthWithUsersProvider`.

## Refactor: shared component/prop cleanup
Prop-driven `ProjectUsersActions`/`ProjectUsersDeleteModal`, responsive `FormFieldsetI18n`, `NO_ROLE` fallback in `usePolicies.formatRole`.

## Testing
Full unit test coverage added/updated across new and touched components.

## Breaking changes
- Old policy API methods removed — any other callers must migrate.
- Role strings now lowercase for these APIs.
- `isUsersProvider` → `isAuthWithUsersProvider`.
- Requires backend `/api/users` + `/api/users/:login/index/:project` support.
  
[Screencast from 2026-07-10 17-15-56.webm](https://github.com/user-attachments/assets/de415d76-d5d1-4e68-ae0f-1bf99cbaeb93)


src/components/Display/DisplayRole.vue
```diff
@@ -1,9 +1,36 @@
...
+import { ROLE, ROLE_ICON_DEFAULT } from '@/enums/roles.js'
+
+const ROLE_ICON = {
+  [ROLE.INSTANCE_ADMIN]: markRaw(IPhPersonSimpleBike),
+  [ROLE.DOMAIN_ADMIN]: markRaw(IPhPersonSimpleRun),
+  [ROLE.PROJECT_ADMIN]: markRaw(IPhPersonSimpleWalk),
+  [ROLE.PROJECT_EDITOR]: markRaw(IPhPersonSimpleHike),
+  [ROLE.PROJECT_MEMBER]: markRaw(IPhPersonSimpleSwim),
+  [ROLE.PROJECT_VISITOR]: markRaw(IPhPersonSimpleTaiChi),
+  [ROLE.NO_ROLE]: ROLE_ICON_DEFAULT,
```
pirhoo: hint: `ROLE` has no `NO_ROLE` member (it's a standalone export), so this key is `undefined` and the intended `NO_ROLE` mapping never registers. It works today thanks to the `?? ROLE_ICON_DEFAULT` fallback, so no change needed, but if you want the entry live, import `NO_ROLE` and key on `[NO_ROLE]`.

src/components/Project/ProjectEntries/ProjectEntriesTableHead.vue
```diff
@@ -57,8 +57,8 @@ const compact = computed(() => {
   />
   <mode-server-only>
     <page-table-th
-      :label="t('projectEntriesTableHead.userRoles')"
-      :icon="IPhUserSquare"
+      :label="t('projectEntriesTableHead.userRole')"
```
pirhoo, line 60: important: this key was renamed from `userRoles` to `userRole` in the template and `en.json`, but `ach/de/es/fr/ja/zh` still define it as `userRoles`, so those locales' values are orphaned and only English survives (ach even keeps an unresolved Crowdin marker). Could we rename the key to `userRole` in the six other locale files too, so translations stay wired up?

src/components/ProjectUsers/ProjectUsersActions.vue
```diff
@@ -0,0 +1,67 @@
...
+    />
+    <button-row-action
+      v-if="!hideDelete"
+      :icon="IPhTrash"
+      :label="deleteActionText"
+      :disabled="disableDelete"
+      @click="showDeleteModal = true"
+    />
+    <project-users-delete-modal
+      v-if="!hideDelete || disableDelete"
```
pirhoo: hint: when `hideDelete && disableDelete` the trigger button is removed but the modal still mounts (`v-if="!hideDelete || disableDelete"`), an instance that can never open. Aligning it with the button (`v-if="!hideDelete"`) would drop the dead mount.

src/components/ProjectUsers/ProjectUsersAdminPromotionModal.vue
```diff
@@ -0,0 +1,49 @@
...
+    </p>
+    <ul class="list-unstyled d-flex flex-column gap-2">
+      <li
+        v-for="promotion in promotions"
+        :key="promotion.login"
+        class="d-flex align-items-center gap-2"
+      >
+        <display-user :value="promotion.login" />
+        <display-role :value="promotion.newRole" />
+      </li>
```
pirhoo, line 46: important: after the uid refactor the `promotions` entries carry `uid`/`newRole` (see `adminPromotions` in ProjectUsersList), so `promotion.login` is undefined here. The admin sees role badges with no usernames and confirms blind, and every `<li>` shares an undefined `:key`.

```suggestion
      <li
        v-for="promotion in promotions"
        :key="promotion.uid"
        class="d-flex align-items-center gap-2"
      >
        <display-user :value="promotion.uid" />
        <display-role :value="promotion.newRole" />
      </li>
```

src/components/ProjectUsers/ProjectUsersCreateModal.vue
```diff
@@ -0,0 +1,236 @@
...
+      email: email.value.trim(),
+      name: name.value.trim(),
+      password: password.value,
+      domain: DEFAULT_DOMAIN,
+      index: props.project,
+      role: ROLE_LOWERCASE[selectedRole.value]
+    })
+    resetForm()
+    modelValue.value = false
+  }
```
pirhoo: important: the try/catch here can't see the create failing. `emit('user:created', ...)` is synchronous, so `resetForm()` runs and the modal closes immediately while the parent's async `onUserCreated` (which does the actual `createUser`/`grantUserRole`) is still in flight. So the modal closes as if it succeeded, and the conflict/error toasts below never fire. Could we move the API calls into the modal (await them here) or have the parent hand back a promise the modal awaits before closing, so the catch is reachable?

src/components/ProjectUsers/ProjectUsersList.vue
```diff
@@ -0,0 +1,221 @@
...
+const emit = defineEmits(['user:deleted'])
+
+const ADMIN_ROLES = new Set([ROLE.PROJECT_ADMIN, ROLE.DOMAIN_ADMIN, ROLE.INSTANCE_ADMIN])
+
+const adminPromotions = computed(() =>
+  Object.entries(pendingChanges.value)
+    .filter(([uid, newRole]) => {
+      const currentRole = props.users.find(u => u.uid === uid)?.role
+      return ADMIN_ROLES.has(newRole) && ROLE_BIT[newRole] > ROLE_BIT[currentRole]
+    })
```
pirhoo, line 57: important: looks like promoting a role-less user straight to admin skips the confirmation modal. `currentRole` is `NO_ROLE` for a user with no access (the fallback in `ProjectViewEditUsers.roleForCurrentProject`), and `NO_ROLE` isn't a key in `ROLE_BIT`, so `ROLE_BIT[newRole] > ROLE_BIT[currentRole]` becomes `n > undefined` which is always false and the promotion goes through silently. Could we default the missing bits to 0?

```suggestion
    .filter(([uid, newRole]) => {
      const currentRole = props.users.find(u => u.uid === uid)?.role
      return ADMIN_ROLES.has(newRole) && (ROLE_BIT[newRole] ?? 0) > (ROLE_BIT[currentRole] ?? 0)
    })
```

src/components/ProjectUsers/ProjectUsersList.vue
```diff
@@ -0,0 +1,221 @@
...
+    .map(([uid, newRole]) => ({ uid, newRole }))
+)
+
+const pendingCount = computed(() => Object.keys(pendingChanges.value).length)
+const hasPendingChanges = computed(() => pendingCount.value > 0)
+
+const sortedUsers = computed(() => orderArrayBy(props.users, [sort.value ?? 'uid'], [order.value]))
```
pirhoo: suggestion: `sortedUsers` re-sorts client-side a page the server already returned sorted and paginated (`retrieveUsers` passes `sort`/`desc`, and changing sort re-fetches). The lodash sort can order case/locale differently from the server and shuffle rows within the page. Could we drop `orderArrayBy` and bind `:items="users"` directly?

src/components/ProjectUsers/ProjectUsersList.vue
```diff
@@ -0,0 +1,221 @@
...
+      )
+    )
+    const revokedUids = entries.filter(([, role]) => role === NO_ROLE).map(([uid]) => uid)
+    entries
+      .filter(([, role]) => role !== NO_ROLE)
+      .forEach(([uid, role]) => {
+        const user = props.users.find(u => u.uid === uid)
+        if (user) user.role = role
+      })
+    pendingChanges.value = {}
```
pirhoo: suggestion: `saveRoles` writes the new role back onto the prop object in place and the parent never refetches after a successful save. It works today only because the parent shares the same object references, and on a partial `Promise.all` failure the catch clears all pending changes with a generic error while some grants already landed, leaving the table out of sync with the server. 

Could we emit a "roles saved" event and let the parent refetch instead of mutating the prop?

src/components/ProjectUsers/ProjectUsersList.vue
```diff
@@ -0,0 +1,221 @@
...
+    toast.success(t('projectViewEdit.users.roleSelect.saveSuccess'))
+    revokedUids.forEach(uid => emit('user:deleted', { uid }))
```
pirhoo: suggestion: this emits `user:deleted` once per revoked uid, and the parent's `onUserDeleted` calls `fetchUsers()` each time, so revoking three roles in one Save fires three concurrent `getUsers` requests whose responses can land out of order. Could we emit a single event (or the revoked list) and refetch once?

src/composables/useAuth.js
```diff
@@ -2,19 +2,22 @@ import { computed, ref, onBeforeMount } from 'vue'
...
 
+const AUTH_MODE_PWD = ['form', 'basic']
+
 export function useAuth() {
   const core = useCore()
 
   const username = ref(null)
   const isBasicAuth = ref(null)
   const isAuthenticated = computed(() => !!username.value)
+  const isAuthWithUsersProvider = computed(() => AUTH_MODE_PWD.includes(core.config.get('auth')))
```
pirhoo, line 14: hint: `isAuthWithUsersProvider` resolves synchronously from `core.config`, but `username` only populates in `onBeforeMount`. While `username` is null, `isCurrentUser(uid)` is false for every row, so the current user's own delete button can briefly render enabled before the async value lands. Gating the delete action on `username.value` being resolved would close the small window.

pirhoo (changes requested):
Impressive work! I've made some suggestions :)

src/components/ProjectUsers/ProjectUsersActions.vue
caro3801: thanks, the condition was wrong but the idea behind was: the modal should be mounted only if the delete button is neither hidden nor disabled. changing to `"!hideDelete && !disableDelete"` ICIJ/datashare-client@34906c808c65853945a1a1248211b35ea5d027e6

src/components/ProjectUsers/ProjectUsersList.vue
caro3801: yes ICIJ/datashare-client@e88ed121e0cbd121cc1d8d4cc03e0fb0bbe9c694

src/components/ProjectUsers/ProjectUsersList.vue
caro3801: yes  ICIJ/datashare-client@98ca9bcf39f32996b4946feadcb9549cce0afb46

src/components/ProjectUsers/ProjectUsersList.vue
caro3801: yes ICIJ/datashare-client@e88ed121e0cbd121cc1d8d4cc03e0fb0bbe9c694

src/components/Display/DisplayRole.vue
caro3801: changed ICIJ/datashare-client@a8f63f926fa23389f32d75f56037f4fdc36964db

src/components/Project/ProjectEntries/ProjectEntriesTableHead.vue
caro3801: updated ICIJ/datashare-client@18ac705a73f168ae9b5f8c61ede14af8cc0fab31

src/components/ProjectUsers/ProjectUsersAdminPromotionModal.vue
caro3801: good catch ! ICIJ/datashare-client@6b94d9ddb0b425b101c0db5ac09c2cc25e79aa04

src/components/ProjectUsers/ProjectUsersList.vue
caro3801: yes ICIJ/datashare-client@36f1bd218140040b66f944ef2fd6f72ca897187d

src/composables/useAuth.js
caro3801: yes! I did not see it thanks ICIJ/datashare-client@e044352c556c716e837c8f629d7ef4d2d2c4d19b

src/components/ProjectUsers/ProjectUsersCreateModal.vue
caro3801: yes I hesitated to have the API inside the modal or not but you're right it makes more sense. Changed ICIJ/datashare-client@9c9dad402f079b359c4a6a8548fb94bc20e5df39

pirhoo (approved):
Thanks for the changes!

PR #445, opened 2026-05-27
feat: async search

pirhoo:
Heavy searches that fan out across many projects can take long enough that the proxy returns an "upstream request timeout" before Elasticsearch answers. This PR moves the main document search onto Elasticsearch's async search API: submit the search, then poll until it is ready so those searches no longer hit that timeout. This is part of icij/datashare#2153.

From the user's point of view nothing changes: the same loading state, then the full results all at once. In-flight searches are now cancelled automatically when a new search supersedes them or when the user leaves the search view.

## Changes:

- feat: run the main document search through async search, with auto cancel (new search or view is left)
- feat: add the async-search client methods and their timing configuration
- fix: stay quiet and correct on cancellation
- test: cover the runner, client methods, store wiring and view cancellation

### DIFF
--- a/src/api/elasticsearch.js
+++ b/src/api/elasticsearch.js
@@ -83,7 +83,7 @@ function handleSearchError(error) {
  * @param {AbortSignal} [signal]
  * @returns {Promise<Object>} The response body.
  */
-function abortableSearchRequest(request, signal) {
+async function abortableSearchRequest(request, signal) {
   const onAbort = () => request.abort?.()
   if (signal) {
     if (signal.aborted) {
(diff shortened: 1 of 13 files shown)


src/api/elasticsearch.js
```diff
@@ -171,7 +230,75 @@ export function datasharePlugin(Client) {
...
+   * @param {Object} options.body - The search body (from buildSearchDocsBody)
+   * @param {string} options.waitForCompletionTimeout - ES duration, e.g. '1s'
+   * @param {string} options.keepAlive - ES duration, e.g. '5m'
+   * @param {AbortSignal} [options.signal] - Cancels the in-flight request.
+   * @returns {Promise<Object>} The async search envelope
+   */
+  Client.prototype.submitAsyncSearch = function ({ index, body, waitForCompletionTimeout, keepAlive, signal }) {
+    const request = this.transport.request({
+      method: 'POST',
+      path: `/${index}/_async_search`,
```
caro3801, line 266:
```suggestion
      path: `/${encodeURIComponent(index)}/_async_search`,
```

caro3801 (approved):
Good implementation and great addition. I learned a lot about async search in ES, thanks!

src/api/elasticsearch.js
pirhoo: Thanks for the suggestion but "index" is a comma-separated list and encodeURIComponent would encode the "," separators to "%2C".

ES decodes it back so it'd still work, but it's semantically wrong (we'd be escaping the delimiter, not the values). And valid ES index names can't contain URL-unsafe characters anyway, so there's nothing to escape :)

PR #450, opened 2026-06-08
feat(SearchToolbar): enable the advanced search toggle button

pirhoo:
As described in icij/datashare#2121, this PR finally activates the advanced search form.

## Changes

* feat: enable the advanced search toggle button in the search toolbar
* feat: reflect an existing query back into the form, mapping `AND`/`OR`/`NOT` (and `+`/`-`) onto the any/all/none word buckets
* feat: add a canonical-AST equivalence check so the form only pre-populates when the regenerated query round-trips without changing meaning — otherwise the modal opens blank instead of silently rewriting the query
* fix: re-emit the user's original query text when an equivalent search is re-submitted, so opening and submitting the modal unchanged never rewrites the search bar
* fix: keep the range-slider label readable in dark mode and centered on the track
* fix: mute only the "e.g." example prefix, keeping the example value in the default body color
* test/docs: cover boolean routing and bare operators, and document the round-trip gate contract

## AI disclosure

The advanced-form round-trip machinery (canonical-AST equivalence, query reflection into the form) and a multi-angle code-review pass were done with Claude Code, with several iterations to harden the round-trip gate and improve code quality.


caro3801 (approved):
huge work on the query parsing! well done

PR #454, opened 2026-06-25
fix: increase hide-margin on download popover to improve the download button hover

caro3801:
Bug:
The download popover disappear when going from hovering the download button to the popover
[Screencast from 2026-06-25 16-41-06.webm](https://github.com/user-attachments/assets/cfe529f5-a8d0-454b-936c-f6b6a4e2bf51)

Fix:
Increase hide-margin on download popover so it covers the button and keep it hovered and visible
[Screencast from 2026-06-25 16-43-33.webm](https://github.com/user-attachments/assets/a7a3ebec-1327-4a18-b8ea-ff39e7ea6d76)

Girl scout:
Fix the race condition in `useUrlParam.js` : detects route changes and discards stale URL parameter updates to prevent cross-route contamination

### DIFF
--- a/src/components/Document/DocumentDownloadPopover/DocumentDownloadPopover.vue
+++ b/src/components/Document/DocumentDownloadPopover/DocumentDownloadPopover.vue
@@ -86,6 +86,7 @@ defineExpose({
     v-model="modelValue"
     hide-header
     class="document-download-popover"
+    :hide-margin="16"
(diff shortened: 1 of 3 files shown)


pirhoo:
I did not know about `hide-margin`, thanks!

pirhoo (approved):

PR #455, opened 2026-06-26
fix(batch-download): deactivate download link when the file is no longer on disk

### LINKED ISSUE #2232: bug: batch download without file on disk should be deactivated
Image

Currently the link to batch downloads with delete a file still has an active link. We should disable that link and show a message bellow the link in the *same fashion and in replacement* than the truncated results message.

In the task items from the API, there is "exists" field under `.items[n].args.batchDownload.exists`.

pirhoo:
Fixes ICIJ/datashare#2232 so a batch download task can complete successfully — recording a result URI — but its zip file can later be removed from disk. The download link was derived purely from the presence of that result URI, so it stayed clickable and 404'd. The backend now reports whether the file is still on disk via `args.batchDownload.exists`.

## Preview 

Screenshot 2026-06-26 at 16-16-40 Batch downloads - Datashare


lschim (approved):

pirhoo:
Thanks @lschim!

PR #458, opened 2026-07-03
fix: disable Download text button when document has no text content

mvanzalu:
This PR disables "Download Text" button when there is no content text in the document

### DIFF
--- a/src/composables/useDocumentDownload.js
+++ b/src/composables/useDocumentDownload.js
@@ -84,11 +84,15 @@ export function useDocumentDownload(document, { immediate = true } = {}) {
     }
   }
 
+  const hasTextContent = computed(() => {
+    return documentRef.value.contentTextLength > 0
+  })
+
   async function downloadTextContent() {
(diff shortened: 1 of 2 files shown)


src/composables/useDocumentDownload.js
```diff
@@ -84,11 +84,17 @@ export function useDocumentDownload(document, { immediate = true } = {}) {
     }
   }
 
-  async function downloadTextContent() {
+  const hasTextContent = ref(true)
+
+  async function fetchTextContentStatus() {
     if (!documentRef.value.content) {
       await documentStore.getContent()
```
pirhoo, line 94: important: this can be very heavy if the text context is big, couldn't we use the `contentTextLength` instead?

pirhoo (changes requested):

src/composables/useDocumentDownload.js
mvanzalu: definitely can

pirhoo (approved):

PR #460, opened 2026-07-07
fix: hide All option in filter list when search query is active

mvanzalu:


pirhoo (changes requested):
important: simple enough but this should also be done to the `FilterTypeProject` which override the "all" template slot.

pirhoo (approved):

PR #463, opened 2026-07-13
fix: reset document selection when exiting selection mode

mvanzalu:
While demonstrating this https://github.com/ICIJ/datashare/issues/2235 we noticed that exiting selection mode does not reinitialize the selected items. As a result, switching in selection mode again (on the same page or on another page) still have the previously selected items.


pirhoo (approved):
That works thanks, though a small comment before the watch would be nice to explain the intent!

PR #464, opened 2026-07-15
fix: enforce user management controls in the project Users tab

caro3801:
## Enforce user management controls in the project Users tab
:warning: merge before https://github.com/ICIJ/datashare/pull/2271
This PR tightens the user CRUD flows added in #443:

- **Restrict user create/delete to instance admins** - both actions are instance-wide on the backend, so the controls are now hidden from project admins (and from non users-provider auth setups).
- **Gate domain/instance admin grants** - `grantUserRole` has no domain/wildcard support yet, so granting those roles would create wrong project-scoped policies. The role dropdown gains `disabledRoles`/`hiddenRoles` props: both options are shown disabled in the users list and hidden in the create-user modal.
- **Fix role display for wildcard policies** - permissions like `*::*` (instance admin) or `default::*` (domain admin) showed as "No role"; the mapping now resolves wildcards and keeps the highest matching role.
- Misc fixes: guard delete/role actions until the current username resolves, await create/grant calls before closing the create modal, remove a dead OAuth branch in the delete modal.

image

### DIFF
--- a/src/components/ProjectUsers/ProjectUsersList.vue
+++ b/src/components/ProjectUsers/ProjectUsersList.vue
@@ -139,12 +139,16 @@ function onUserDeleted({ uid }) {
   emit('user:deleted', { uid })
 }
 const { username, isUsernameResolved, isAuthWithUsersProvider } = useAuth()
-const { isInstanceAdmin } = usePolicies()
+const { isInstanceAdmin, getRoleByProject, hasRole } = usePolicies()
 // Deleting a user account removes it from every project (see the delete modal's warning) and the
 // backend requires INSTANCE_ADMIN, so hide the action from project admins who are not instance admins.
 const canDeleteUsers = computed(() => isAuthWithUsersProvider.value && isInstanceAdmin())
+const viewerRole = computed(() => getRoleByProject(props.project))
 function isCurrentUser(uid) {
   return !isUsernameResolved.value || username.value === uid
+}
+function outranksViewer(role) {
+  return !hasRole(viewerRole.value, role)
 }
 
 defineExpose({ pendingChanges, saving, showAdminModal, saveRoles, cancelChanges, onSaveClicked })
@@ -170,7 +174,7 @@ defineExpose({ pendingChanges, saving, showAdminModal, saveRoles, cancelChanges,
       </template>
       <template #cell(role)="{ item }">
         <project-users-role-dropdown
-          :disabled="isCurrentUser(item.uid)"
+          :disabled="isCurrentUser(item.uid) || outranksViewer(item.role)"
           :model-value="pendingChanges[item.uid] ?? item.role"
           :dirty="!!pendingChanges[item.uid]"
           :project="project"
--- a/src/components/ProjectUsers/ProjectUsersList.vue
+++ b/src/components/ProjectUsers/ProjectUsersList.vue
@@ -174,6 +174,7 @@ defineExpose({ pendingChanges, saving, showAdminModal, saveRoles, cancelChanges,
           :model-value="pendingChanges[item.uid] ?? item.role"
           :dirty="!!pendingChanges[item.uid]"
           :project="project"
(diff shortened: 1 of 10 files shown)


src/components/ProjectUsers/ProjectUsersList.vue
pirhoo: important: looks like a project admin can still act on a higher-privileged admin's row. `disabled-roles` only greys out the DOMAIN_ADMIN/INSTANCE_ADMIN options, which a project admin never sees anyway, so when such an admin now appears in the list (via the new wildcard resolution) the lower roles stay clickable and Save calls `grantUserRole` against them. Could we disable the whole dropdown when the row's own role outranks what the viewer may assign? The backend is the real backstop, but the control should not be actionable here.

```suggestion
          :disabled="isCurrentUser(item.uid) || [ROLE.DOMAIN_ADMIN, ROLE.INSTANCE_ADMIN].includes(pendingChanges[item.uid] ?? item.role)"
```

This obviously would be better in dedicated function.

src/views/Project/ProjectView/ProjectViewEdit/ProjectViewEditUsers.vue
```diff
@@ -64,8 +69,14 @@ watch(query, (value) => {
...
-  const permission = (permissions ?? []).find(({ v2 }) => v2 === `${DEFAULT_DOMAIN}::${props.name}`)
-  return permission?.v1 ?? NO_ROLE
+  // Permissions are "<domain>::<project>" pairs where either side can be the "*" wildcard
+  // (e.g. "*::*" for instance admins, "default::*" for domain admins).
+  const matching = (permissions ?? []).filter(({ v2 }) => {
+    const [domain, project] = String(v2).split('::')
+    return (domain === DEFAULT_DOMAIN || domain === '*') && (project === props.name || project === '*')
+  })
+  if (!matching.length) return NO_ROLE
+  return getHighestRoleFromList(matching.map(({ v1 }) => ({ role: v1 })))
```
pirhoo: important: `getHighestRoleFromList` resolves to `PROJECT_MEMBER` (via `resolveHighestRole`'s `DEFAULT_ROLE` fallback) whenever the combined bitmask is 0, so a matched permission whose `v1` is not a `ROLE_BIT` key (an empty string, a revoked/`NO_ROLE` sentinel, any casing the backend emits) is shown as Member instead of No role, overstating access. The `!matching.length` guard misses this because the match exists. Could we keep only recognized roles and fall back to `NO_ROLE`? Needs adding `ROLE_BIT` to the roles import.

```suggestion
  const matchesCurrentProject = ({ v2 }) => {
    const [domain, project] = String(v2).split('::')
    const domainMatches = domain === DEFAULT_DOMAIN || domain === '*'
    const projectMatches = project === props.name || project === '*'
    return domainMatches && projectMatches
  }
  const matchingPermissions = (permissions ?? []).filter(matchesCurrentProject)
  const knownRoles = matchingPermissions.map(({ v1 }) => ({ role: v1 })).filter(({ role }) => ROLE_BIT[role])
  if (!knownRoles.length) {
    return NO_ROLE
  }
  return getHighestRoleFromList(knownRoles)
```

pirhoo (changes requested):
Thanks for the enhancements, I suggested two changes :)

src/components/ProjectUsers/ProjectUsersList.vue
caro3801: yes, more user friendly thanks ICIJ/datashare-client@257ecd20c

src/views/Project/ProjectView/ProjectViewEdit/ProjectViewEditUsers.vue
caro3801: yes good catch ICIJ/datashare-client@6f3c00524

caro3801:
thanks for the review !

pirhoo (approved):

PR #466, opened 2026-07-23
refactor: use murmur tree shaking

caro3801:
Refactor `@icij/murmur` usage to enable tree-shaking: 
* named imports for `config/i18n` instead of the default export, 
* direct component imports in SFCs instead of global plugin registration, 
* and lazy-load of the rarely-used Murmur components in `getMurmurComponent` 

**🍪 Girl Scout side quest** - while testing the refactor, found and fixed console warnings, each with a regression test:
- `TaskBoardListEntry`: `icon` prop still typed as `String` after `TASK_NAME_ICON` moved to `unplugin-icons` component objects
- `SearchHistoryList`/`SearchSavedList`: custom breadcrumb slots omitted `label`/`context` on overflow entries, so overflowed items resolved to a `null` route
- `FilterTypeDateRange`: stray `size="1em"` (an icon-sized value) accidentally landed on `FormControlDateRange`'s `size` prop, which only accepts `sm`/`md`/`lg`
- `BatchSearchCardDetails`: typo (`phraseMatch` vs `phraseMatches`) picked the wrong proximity/fuzziness entry
- add guards to prevent access to `task.entities` routes in server mode

### DIFF
--- a/src/api/resources/Document.js
+++ b/src/api/resources/Document.js
@@ -1,6 +1,6 @@
 import { compact, endsWith, filter, find, get, keys, last, pick, startsWith, trim } from 'lodash'
 import { markRaw } from 'vue'
-import Murmur from '@icij/murmur'
+import { config } from '@icij/murmur'
 import dayjs from 'dayjs'
 import { extname } from 'path'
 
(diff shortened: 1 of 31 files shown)


pirhoo (approved):

pirhoo:
Nice and tidy 🧹

PR #470, opened 2026-07-28
perf(build): add manualChunks vendor-chunking strategy

caro3801:
## Summary
- Adds `build.rollupOptions.manualChunks` in `vite.config.js`, grouping heavy, eagerly-reachable vendor deps (search/ES client, charts/calendar, UI kit, core framework, lodash) into their own cache-friendly chunks instead of bundling with whatever route imported them first.
- `xlsx` (spreadsheet export) also gets its own chunk since it's feature-specific.
- No catch-all chunk and no markdown grouping: anything not explicitly matched keeps Rollup's default placement, so already-lazy code (the markdown stack, PDF/TIFF/image viewers) stays lazy instead of being pulled into the eager preload set.

## Review history
An earlier version of this PR had a catch-all `vendor` chunk and a `vendor-markdown` group. Both were eagerly reachable from the entry, so they defeated the point: `pdfjs-dist`/`tiff`/`image-js` and the markdown stack got preloaded on first paint instead of staying in their lazy document-viewer chunks, a real first-load regression (measured 1,160 kB gzip vs. 791 kB on `main`). Fixed per review by dropping both groups, correcting the header comment's claim that grouping "keeps things out of the critical path" (`manualChunks` only affects caching, not reachability), and fixing the package-name regex to scope from the last `node_modules` segment.

## Measured impact (real browser, real backend, `performance.getEntriesByType('resource')`, hard-reload to bypass stale cache)

**Initial page load, total JS transferred:**
| | `main` (no manualChunks) | this branch |
|---|---|---|
| Total JS transferred | 867,489 B (~847 kB) | 840,856 B (~821 kB) |
| Largest chunk | one 745,366 B monolith | `index.js` 241,451 B + `vendor-ui` 159 kB + `vendor-charts` 144 kB + `vendor-search` 103 kB + `vendor-framework` 95 kB + `vendor-lodash` 65 kB |

First-load bytes are roughly flat (~3% smaller), not a regression, confirming the fixed version behaves as the reviewer expected: `manualChunks` doesn't meaningfully shrink first load since these deps are eagerly reachable either way.

**The actual benefit is repeat-deploy caching**, not first load: with `main`'s single monolith, any app-code change forces a full 745 kB re-download for every user. With this branch, app code lives in `index.js` (241 kB) while 567 kB of vendor code sits in independently-hashed chunks that stay cached across app-only deploys, roughly a 68% smaller re-download for that common case.

## Test plan
- [x] Verified via `vite preview` + real browser against the live local backend: project list, search, and document viewer (spreadsheet + image) all load cleanly, zero console exceptions, all requests 200
- [x] Confirmed lazy chunks (e.g. `DocumentViewerLegacySpreadsheet`) only fetch on demand, not on initial page load
- [x] Confirmed no circular-chunk-dependency runtime errors (an initial `vendor-viewer` grouping for pdfjs-dist/image-js/tiff threw at runtime; reverted that specific grouping)
- [x] Measured initial-load JS transfer bytes against `main` baseline (see above)

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -128,33 +128,44 @@ export default ({ mode }) => {
           /**
            * No vendor-chunking strategy existed before this: heavy,
            * rarely-changing dependencies got bundled into whichever route
-           * chunk first imported them (see log.md 2026-07-24/27 bundle
-           * analysis). Grouping them into their own chunks lets the browser
-           * cache them independently of app-code deploys and keeps them out
-           * of the search route's critical path.
+           * chunk first imported them. Grouping the ones that are eagerly
+           * reachable anyway into their own chunks lets the browser cache
+           * them independently of app-code deploys.
+           *
+           * This does not defer anything: a group is downloaded on first
+           * paint as soon as one member is statically reachable from the
+           * entry. Deferring needs a dynamic import at the call site, not a
+           * chunk name here.
            */
           manualChunks(id) {
             if (!id.includes('node_modules')) return
 
-            const match = id.match(/node_modules\/((?:@[^/]+\/)?[^/]+)/)
+            const scope = id.slice(id.lastIndexOf('node_modules/') + 'node_modules/'.length)
+            const match = scope.match(/^((?:@[^/]+\/)?[^/]+)/)
             const pkg = match?.[1]
             if (!pkg) return
 
-            // Search/ES client stack — single biggest dependency (log.md:
-            // elasticsearch-browser alone is ~1.6MB raw).
+            // Search/ES client stack — elasticsearch-browser alone is
+            // ~1.6 MB raw, the single biggest dependency in the app.
             if (['elasticsearch-browser', 'bodybuilder', 'lucene'].includes(pkg)) {
               return 'vendor-search'
             }
 
-            // Charting/geo/calendar — mostly project-overview widgets, not
-            // needed on the search route itself.
+            // Charting/geo/calendar. Despite the name, this is not deferred
+            // to project-overview: ColumnChartPicker.vue imports d3 directly
+            // and is statically reachable from the search route itself (via
+            // FilterTypeDateRange.vue -> FilterDateRange.js ->
+            // store/filters/index.js -> search.js), which the eager store
+            // barrel in Core.js pulls in on every page..
+            // Isolating it here still lets it cache independently of
+            // app-code deploys; actually deferring it needs breaking that
+            // eager import chain, not a chunk name.
             if (pkg === 'd3' || pkg.startsWith('d3-') || pkg === 'v-calendar' || pkg.startsWith('topojson')) {
               return 'vendor-charts'
             }
 
             // UI kit — bootstrap-vue-next and @icij/murmur share a large
-            // amount of code (see log.md 2026-07-27) and are used across
-            // nearly every view.
+            // amount of code and are used across nearly every view.
             if (['bootstrap-vue-next', '@icij/murmur', 'bootstrap'].includes(pkg) || pkg.startsWith('@floating-ui')) {
               return 'vendor-ui'
             }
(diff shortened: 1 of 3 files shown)


vite.config.js
```diff
@@ -114,6 +121,82 @@ export default ({ mode }) => {
...
+            }
+
+            // lodash + the standalone lodash.* packages (log.md: 885 kB raw,
+            // duplicated logic across lodash and lodash.merge/clonedeep/unset)
+            // — isolate so it caches independently of unrelated vendor code.
+            if (pkg === 'lodash' || pkg.startsWith('lodash.')) {
+              return 'vendor-lodash'
+            }
+
+            return 'vendor'
```
pirhoo: crucial: this catch-all is what makes the branch a net first-load regression. `vendor` is statically reachable from the entry, so everything that falls through here gets `modulepreload`ed on page one, including `pdfjs-dist` and `tiff`/`image-js` that `main` kept in the lazily-imported `DocumentViewerPdfPage` (416 kB) and `DocumentViewerTiff` (265 kB) chunks. Measured first-load goes from 791 kB gzip on `main` to 1160 kB here; dropping the line puts it at 858 kB and rollup's default placement returns those deps to their lazy route chunks.

```suggestion
```

vite.config.js
```diff
@@ -114,6 +121,82 @@ export default ({ mode }) => {
...
+      rollupOptions: {
+        output: {
+          /**
+           * No vendor-chunking strategy existed before this: heavy,
+           * rarely-changing dependencies got bundled into whichever route
+           * chunk first imported them (see log.md 2026-07-24/27 bundle
+           * analysis). Grouping them into their own chunks lets the browser
+           * cache them independently of app-code deploys and keeps them out
+           * of the search route's critical path.
+           */
```
pirhoo, line 139: important: the caching half of this holds up, but "keeps them out of the search route's critical
path" does not: `manualChunks` only changes which file a module lands in, never whether it is
eagerly reachable. `dist/index.html` preloads every group except `vendor-export`, and it cannot be
otherwise for `vendor-charts` (`v-calendar` is registered as a global plugin in `src/core/Core.js`,
and `d3`'s `range` comes in via `src/composables/useRangeSteps.js` for the range filters) or for
`vendor-search`.

```suggestion
          /**
           * No vendor-chunking strategy existed before this: heavy,
           * rarely-changing dependencies got bundled into whichever route
           * chunk first imported them. Grouping the ones that are eagerly
           * reachable anyway into their own chunks lets the browser cache
           * them independently of app-code deploys.
           *
           * This does not defer anything: a group is downloaded on first
           * paint as soon as one member is statically reachable from the
           * entry. Deferring needs a dynamic import at the call site, not a
           * chunk name here.
           */
```

vite.config.js
```diff
@@ -114,6 +121,82 @@ export default ({ mode }) => {
...
+            // Spreadsheet export — feature-specific, not needed app-wide.
+            if (pkg === 'xlsx') {
+              return 'vendor-export'
+            }
+
+            // Markdown rendering stack (micromark/mdast/unified/hast) — used
+            // for rendering release-notes/help content, not core navigation.
+            if (['micromark', 'unified', 'vfile', 'property-information'].includes(pkg) || pkg.startsWith('micromark-') || pkg.startsWith('mdast-') || pkg.startsWith('hast-') || pkg.startsWith('unist-') || pkg.startsWith('remark-')) {
+              return 'vendor-markdown'
+            }
```
pirhoo: important: the markdown stack only enters the graph through `src/utils/markdown.js`, which is reached from `DocumentViewerMarkdown.vue`, itself behind the `defineAsyncComponent` in `DocumentViewTabsViewer.vue`. Rollup already kept it lazy on `main`, naming a group for it pulls it into the preload set instead, so we're better off leaving it alone.

```suggestion
```

vite.config.js
```diff
@@ -114,6 +121,82 @@ export default ({ mode }) => {
...
+           * rarely-changing dependencies got bundled into whichever route
+           * chunk first imported them (see log.md 2026-07-24/27 bundle
+           * analysis). Grouping them into their own chunks lets the browser
+           * cache them independently of app-code deploys and keeps them out
+           * of the search route's critical path.
+           */
+          manualChunks(id) {
+            if (!id.includes('node_modules')) return
+
+            const match = id.match(/node_modules\/((?:@[^/]+\/)?[^/]+)/)
```
pirhoo: suggestion: the match lands on the first `node_modules` segment, so a non-hoisted copy at
`node_modules/foo/node_modules/lodash/...` gets attributed to `foo` and skips the `vendor-lodash`
group. Nothing in the current tree hits this (the nested copies are storybook-only), so it's cheap
insurance rather than a live bug.

```suggestion
            const scope = id.slice(id.lastIndexOf('node_modules/') + 'node_modules/'.length)
            const match = scope.match(/^((?:@[^/]+\/)?[^/]+)/)
```

vite.config.js
pirhoo, line 1: hint: the five `log.md` citations point at a file that isn't in the repo, so the next reader can't follow the reasoning back to its numbers. Worth either inlining the figure that mattered or dropping the reference.

pirhoo (changes requested):

caro3801:
Thanks for the review, all 5 points addressed in fa212dcb2:

1. Dropped the catch-all `vendor` chunk. pdfjs-dist/tiff/image-js are back in their lazy document-viewer chunks.
2. Replaced the header comment with accurate framing: manualChunks only affects caching, not reachability/deferral.
3. Dropped the `vendor-markdown` group so that stack stays lazy behind `DocumentViewerMarkdown`.
4. Fixed the package regex to scope from the last `node_modules` segment.
5. Removed the log.md citations and inlined the relevant figures directly in the comments.

Also measured the actual before/after against `main`: initial JS transfer is now 821 kB vs 847 kB baseline (was 1.16 MB with the catch-all bug), and verified live against the real backend (search, spreadsheet viewer, image viewer all load correctly, lazy chunks only fetch on demand, no console errors). Numbers and methodology are in the updated PR description.

pirhoo (approved):
Wonderful, thanks!

PR #471, opened 2026-07-28
fix(recommendations): gate server-only fetches by mode

caro3801:
## Summary
- `useSearchFilter.js`'s `refreshRecommendedBy()` ran on every `refreshSearch()` regardless of mode, only incidentally avoided a fetch via `getDocumentsRecommendedBy`'s `indices.length + users.length > 1` check. Gated behind `useMode(core).isServer`.
- `useDocument.js`'s `fetchDocument()` unconditionally called `documentStore.getRecommendationsByDocuments(...)`, hitting a server-only endpoint (404 in local mode). Same guard applied.
- `ProjectDeletionModal.vue`'s `onBeforeMount` unconditionally called `fetchRecommendationsCount()`, not mode-gated even though project deletion is a LOCAL/EMBEDDED feature too. Same guard applied.
- Confirmed `FilterTypeRecommendedBy.vue` is already safe: it fetches unconditionally in its own code, but only mounts when `SearchFilters.vue` has already filtered filters by `filter.modes.some(isMode)`, so it never mounts outside server mode.

## Test plan
- [x] `useSearchFilter.spec.js`: 3 new tests, no fetch in LOCAL/EMBEDDED mode, fetch in SERVER mode
- [x] `DocumentView.spec.js`: 2 new tests, server called / local not called
- [x] `ProjectDeletionModal.spec.js`: 2 new tests, same pattern

### DIFF
--- a/src/components/Project/ProjectDeletionModal.vue
+++ b/src/components/Project/ProjectDeletionModal.vue
@@ -9,7 +9,6 @@ import AppModal from '@/components/AppModal/AppModal'
 import DisplayNumber from '@/components/Display/DisplayNumber'
 import ProjectLabel from '@/components/Project/ProjectLabel'
 import { useProjectMetrics } from '@/composables/useProjectMetrics'
-import { useMode } from '@/composables/useMode'
 import { useWait } from '@/composables/useWait'
 import { useCore } from '@/composables/useCore'
 import { useToast } from '@/composables/useToast'
@@ -30,7 +29,6 @@ const recommendationsCount = ref(0)
 
 const { waitFor, loaderId } = useWait()
 const core = useCore()
-const { isServer } = useMode(core)
 const { toast } = useToast()
 const { fetchDocumentsCount, fetchTagsCount, fetchRecommendationsCount } = useProjectMetrics(toRef(props, 'project'))
 
@@ -54,10 +52,7 @@ onBeforeMount(
   waitFor(async () => {
     documentsCount.value = await fetchDocumentsCount()
     tagsCount.value = await fetchTagsCount()
-    // Recommendations are a server-mode-only feature
-    if (isServer.value) {
-      recommendationsCount.value = await fetchRecommendationsCount()
-    }
+    recommendationsCount.value = await fetchRecommendationsCount()
   })
 )
 </script>
--- a/src/components/Project/ProjectDeletionModal.vue
+++ b/src/components/Project/ProjectDeletionModal.vue
@@ -9,6 +9,7 @@ import AppModal from '@/components/AppModal/AppModal'
 import DisplayNumber from '@/components/Display/DisplayNumber'
 import ProjectLabel from '@/components/Project/ProjectLabel'
 import { useProjectMetrics } from '@/composables/useProjectMetrics'
+import { useMode } from '@/composables/useMode'
 import { useWait } from '@/composables/useWait'
 import { useCore } from '@/composables/useCore'
 import { useToast } from '@/composables/useToast'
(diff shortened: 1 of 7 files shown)


mvanzalu (approved):

tests/unit/specs/composables/useSearchFilter.spec.js
```diff
@@ -849,3 +850,67 @@ describe('useSearchFilter composable', () => {
     })
   })
 })
+
+describe('useSearchFilter refreshSearch with recommendations gated to server mode', () => {
```
pirhoo: suggestion: this new top-level describe re-creates the `beforeEach` and the `mountComposable` helper that the `useSearchFilter composable` describe above already has, so the two copies will drift. Could we nest it in that describe instead, hoisting its local `core` into the existing `let plugins, searchStore` and adding only the `recommendedStore` spy in a nested `beforeEach`?

src/composables/useSearchFilter.js
```diff
@@ -343,6 +345,12 @@ export function useSearchFilter() {
   }
 
   function refreshRecommendedBy() {
+    // Recommendations are a server-mode-only feature (see widget
+    // registration's `modes: [MODE_NAME.SERVER]` in store/widgets/index.js).
+    // Skip the fetch entirely in local/embedded mode instead of relying on
+    // getDocumentsRecommendedBy's incidental `indices.length + users.length
+    // > 1` guard, which doesn't actually check mode.
```
pirhoo: suggestion: the pointer here sends the reader to the widget registration, but what actually gates this code path is the filter itself, registered with `modes: [MODE_NAME.SERVER]` in `store/filters/index.js`. Same note applies to the copy of this comment in `useDocument.js`.

```suggestion
    // Recommendations are a server-mode-only feature (see the
    // `modes: [MODE_NAME.SERVER]` on FilterRecommendedBy in
    // store/filters/index.js). Skip the fetch entirely in local/embedded mode
    // instead of relying on getDocumentsRecommendedBy's incidental
    // `indices.length + users.length > 1` guard, which doesn't check mode.
```

src/components/Project/ProjectDeletionModal.vue
```diff
@@ -52,7 +54,10 @@ onBeforeMount(
   waitFor(async () => {
     documentsCount.value = await fetchDocumentsCount()
     tagsCount.value = await fetchTagsCount()
-    recommendationsCount.value = await fetchRecommendationsCount()
+    // Recommendations are a server-mode-only feature
+    if (isServer.value) {
+      recommendationsCount.value = await fetchRecommendationsCount()
+    }
```
pirhoo: hint: `fetchRecommendationsCount` is the only server-only metric in `useProjectMetrics`, so the guard could live in the composable (return 0 when not in server mode) and this component would not need `useMode` at all.

pirhoo (changes requested):
Only suggestions, the overalls changes are good!

tests/unit/specs/composables/useSearchFilter.spec.js
caro3801: yes makes sense

src/components/Project/ProjectDeletionModal.vue
caro3801: yes! moved the mode guard into useProjectMetrics's fetchRecommendationsCount (returns 0 when not server mode)

tests/unit/specs/composables/useSearchFilter.spec.js
pirhoo: Thanks!

src/components/Project/ProjectDeletionModal.vue
pirhoo: I think we still need to remove the if statement here (and the import of `useMode`and `isServer`).

pirhoo (approved):

PR #472, opened 2026-07-28
perf(recommendations): fetch translation status only when the download popover opens

caro3801:
## Summary
- Translation status was fetched eagerly on visibility for every document's download button, even though only the download popover (opened on click) actually reads it.
- Split `useDocumentDownload`'s combined `fetchStatuses` so `DocumentActionsGroupEntryDownload` (the button) fetches only `fetchDownloadStatus` on scroll-into-view, leaving `fetchTranslationStatus` to the popover's own on-open fetch (`whenever(activated, fetchStatuses, { once: true })`).
- `documentDownloadStore` memoizes `fetchTranslationStatus` per `index:id` key, so re-opening an already-fetched document's popover costs nothing.

## Test plan
- [x] `DocumentDownloadPopover.spec.js`, `useDocumentDownload.spec.js`, `DocumentActionsGroupEntryDownload.spec.js` — 22 tests passing

### DIFF
--- a/src/store/modules/documentDownload.js
+++ b/src/store/modules/documentDownload.js
@@ -70,15 +70,21 @@ export const useDocumentDownloadStore = defineStore('documentDownload', () => {
   const fetchTranslationStatus = async ({ index, id, routing }) => {
     const key = translationCacheKey({ index, id })
     if (key in translationsFor) return translationsFor[key]
-    try {
-      const _source = 'content_translated.target_language'
-      const data = await api.elasticsearch.getSource({ index, id, routing, _source })
-      translationsFor[key] = extractTranslations(data)
+    fetchPromises[key] ??= (async () => {
+      try {
+        const _source = 'content_translated.target_language'
+        const data = await api.elasticsearch.getSource({ index, id, routing, _source })
+        translationsFor[key] = extractTranslations(data)
+      }
+      catch {
+        return []
+      }
+      finally {
+        delete fetchPromises[key]
+      }
       return translationsFor[key]
-    }
-    catch {
-      return []
-    }
+    })()
+    return fetchPromises[key]
   }
 
   return {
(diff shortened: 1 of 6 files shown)


src/store/modules/documentDownload.js
```diff
@@ -49,5 +58,32 @@ export const useDocumentDownloadStore = defineStore('documentDownload', () => {
...
+    try {
+      const _source = 'content_translated.target_language'
+      const data = await api.elasticsearch.getSource({ index, id, routing, _source })
+      translationsFor[key] = extractTranslations(data)
+    }
+    catch {
+      translationsFor[key] = []
+    }
+    return translationsFor[key]
+  }
```
pirhoo, line 88: important: looks like a failed probe gets cached as permanently "no translations": the catch writes `[]` into `translationsFor`, and the `key in translationsFor` guard then short-circuits every later call. A transient 5xx or network blip would hide the translated-content button for that document for the rest of the session, where the old per-composable `watchEffect` retried on the next mount. Could we only cache successes?

```suggestion
  const fetchTranslationStatus = async ({ index, id, routing }) => {
    const key = translationCacheKey({ index, id })
    if (key in translationsFor) return translationsFor[key]
    try {
      const _source = 'content_translated.target_language'
      const data = await api.elasticsearch.getSource({ index, id, routing, _source })
      translationsFor[key] = extractTranslations(data)
      return translationsFor[key]
    }
    catch {
      return []
    }
  }
```

pirhoo (changes requested):

src/store/modules/documentDownload.js
caro3801: good catch, thanks, I applied your suggested diff as-is

caro3801:
Also added same in-flight de-dup that `fetchDocumentStatus` uses (`fetchPromises[key] ??=` ...)  to `fetchTranslationStatus` so two callers probing the same document before the first resolves will fire only one request [6a0e725](https://github.com/ICIJ/datashare-client/pull/472/commits/6a0e725ed577a65ef26a7cfe496e98a5f783b627)

pirhoo (approved):

PR #475, opened 2026-07-29
fix: redirect users to their requested page after login in server mode

pirhoo:
In server mode, an unauthenticated user opening a deep link (shared document, search URL) was redirected to login but landed on the default page after logging in, for both form auth and OAuth. The intended destination is now stored before redirecting to login (including from the boot 401 path) and restored, with query and hash, on the first authenticated navigation.

- fix(store): only accept relative paths as redirect after login
- fix(store): reject protocol-relative paths as redirect after login
- fix(router): preserve query and hash in redirect after login
- fix(router): check authentication before user role in guards
- fix(core): store requested url before redirecting to login on boot 401
- test(core): await boot failure navigation deterministically
- fix(core): capture hash route for redirect after boot 401

### DIFF
--- a/src/store/modules/app.js
+++ b/src/store/modules/app.js
@@ -188,6 +188,9 @@ export const useAppStore = defineStore(
       else if (path.startsWith('/') && !path.startsWith('//') && !path.startsWith('/\\') && !path.startsWith('/login')) {
         redirectAfterLogin.value = path
       }
+      else { 
+          redirectAfterLogin.value = null
+      }
     }
 
     /**
--- a/src/store/modules/app.js
+++ b/src/store/modules/app.js
@@ -175,12 +175,17 @@ export const useAppStore = defineStore(
     }
 
     /**
-     * Sets the redirect path to be used after login.
+     * Sets the redirect path to be used after login. Only relative paths
+     * (starting with `/`) are accepted, and never the login page itself.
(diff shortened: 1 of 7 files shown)


src/store/modules/app.js
```diff
@@ -175,12 +175,17 @@ export const useAppStore = defineStore(
...
      * @param {string|null} [path=null] - Redirect path.
      */
     const setRedirectAfterLogin = (path = null) => {
-      if (!path || !path.startsWith('/login')) {
+      if (!path) {
+        redirectAfterLogin.value = null
+      }
+      else if (path.startsWith('/') && !path.startsWith('//') && !path.startsWith('/\\') && !path.startsWith('/login')) {
         redirectAfterLogin.value = path
       }
```
caro3801, line 191: suggestion:  when path fails validation (not /-prefixed, or //, /\, /login), `setRedirectAfterLogin` falls through silently and leaves the old persisted `redirectAfterLogin.value` in place instead of clearing it. 

```suggestion
      }
      else { 
          redirectAfterLogin.value = null
      }
```

caro3801 (approved):
thanks! I did not realize it caused that much friction until it was fixed!

src/store/modules/app.js
pirhoo: Good idea, thanks!

PR #479, opened 2026-08-07
perf: tree-shake lodash/d3, lazy-load optional search components, fix search-stuck-after-History bug

caro3801:
Search-page performance work: reduces JS bundled/eagerly-loaded on every page load by tree-shaking lodash and d3, lazy-loading components that only matter for optional facets/widgets, and adding a `manualChunks` vendor split. Also fixes a live bug where the search view got stuck on skeleton placeholders after navigating back to "Documents" from "History".

### Changes

**Bundle size / lazy-loading**
- `082195cdd` — tree-shake lodash via `lodash/<fn>` deep imports instead of the `lodash-es` barrel; drops the `lodash-es` dependency entirely.
- `99dcdaeca` — lazy-load the `SearchAdvancedModal` chunk until first opened.
- `ac7ce699e` — lazy-load `fuse.js` in `SearchFilters` until first keystroke.
- `4c975cf2a` — defer `ColumnChartPicker`/d3 load in the date-range filter until it's opened.
- `37a459149` — lazy-load widget Vue components referenced from the `store/widgets` barrel.
- `24ec73c7b` — stop eagerly installing the `v-calendar` plugin at boot.
- `b5b7e1fce` — lazy-load the optional `FilterType*` components (`FilterTypeDateRange`, `FilterTypePath`, `FilterTypeProject`, `FilterTypeRecommendedBy`, `FilterTypeStarred`); each is only relevant to a project enabling that specific facet.
- `d303d57bb` — replace bare/namespace `d3` barrel imports in `ColumnChartPicker.vue`, `WidgetDocumentsByCreationDate.vue`, and `useRangeSteps.js` with per-submodule imports (`d3-array`, `d3-scale`, `d3-time`, `d3-time-format`); drops the `d3` dependency; adds an eslint rule blocking bare `d3`/`lodash` imports repo-wide.
- `be8aa8b29` — dedupe a duplicate `d3-format` resolution (yarn kept two copies since `d3-scale`'s `"1 - 3"` range and `@icij/murmur`'s `"^3.1.0"` range didn't string-match); pinned via `resolutions`. Also hardens a consistently-failing cold-dynamic-import test timeout in `FilterTypeDateRange.spec.js`.

**Bug fix**
- `873c4640e` — prevent cancelled/superseded queries from being counted as applied. `resetForRouteChange()` unconditionally forced `isReady = false` on every route entry, but nothing flipped it back when the route-query watchers correctly skipped re-fetching an already-satisfied query. Now `isReady` only resets when the query actually changed.

**Dependency bumps**
- `f711d9a89`, `bd0678eeb`, `b36406ca4` — bump `@icij/murmur` 5.1.0 → 5.1.1 → 5.1.2 (5.1.2 fixes an upstream ES-build issue that broke `yarn vitest`).

### Impact

- Main `index` chunk: **2,617 kB → ~602 kB raw**.
- `vendor-lodash` chunk: **180 kB → 106.5 kB raw**.
- `vendor-charts` chunk: small reduction (254.7 kB → 250.17 kB raw); the bulk remains unused `@icij/murmur` components (`LineChart`, `ChoroplethMap`, `SymbolMap`) pulled in via murmur's single barrel — not fixable from this repo alone, left open.

### Not done here

- `elasticsearch-browser` (~1.6 MB, the single biggest bundle item) — untouched, needs its own effort.

### Testing

Targeted spec files per change (not the full suite, per repo convention) — `search.spec.js`, `useSearchFilter.spec.js`, `views/Search/*`, `ColumnChartPicker.spec.js`, `useRangeSteps.spec.js`, `WidgetDocumentsByCreationDate*.spec.js`, `FilterTypeDateRange.spec.js`, `SearchToolbar.spec.js`, `WidgetNested.spec.js` — all passing. `yarn build` verified after bundle-shape changes; `yarn lint --fix` clean throughout. The search-stuck fix was TDD'd against a failing test reproducing the exact race.


src/components/Search/SearchToolbar/SearchToolbar.vue
```diff
@@ -95,6 +106,7 @@ function handleAdvancedSearch({ query, field }) {
       />
     </div>
     <search-advanced-modal
+      v-if="hasOpenedAdvancedSearch"
```
pirhoo, line 109: crucial: the modal now mounts with `modelValue` already `true`, so `whenever(isVisible, handleOpen)` in `SearchAdvancedModal` never fires on a first open: empty form, no autofocus, and submitting overwrites the active query. Could we add `{ immediate: true }` to that `whenever`?

src/store/modules/search.js
```diff
@@ -274,9 +272,13 @@ export const useSearchStore = defineSuffixedStore('search', () => {
...
    */
-  function resetForRouteChange() {
+  function resetForRouteChange(routeQuery = {}) {
     from.value = 0
-    isReady.value = false
+    // Only force loading state for a genuinely new query, or a same-query
+    // route re-entry (e.g. History → Documents) leaves it stuck forever.
+    if (!sameAppliedQuery(routeQuery, ['from'])) {
+      isReady.value = false
+    }
```
pirhoo, line 282: important: this rarely matches a real URL, so the stuck-loading case it targets survives: `route.query` carries a `stamp` that never equals `queryStamp`, and a single-valued filter arrives as a scalar where `lastAppliedQuery` holds an array. `sameAppliedQuery({})` is also vacuously true, so entering the route with no params keeps `isReady` true over a wiped `q`. Below fixes the stamp and empty cases; filter values still need a `castArray` before comparing.

```suggestion
    // Only force loading state for a genuinely new query, or a same-query
    // route re-entry (e.g. History → Documents) leaves it stuck forever.
    const sameQuery = Object.keys(routeQuery).length && sameAppliedQuery(routeQuery, ['from', 'stamp'])
    if (!sameQuery) {
      isReady.value = false
    }
```

vite.config.js
```diff
@@ -151,15 +151,13 @@ export default ({ mode }) => {
...
-            // eager import chain, not a chunk name.
+            // Charting/geo/calendar. d3 (via ColumnChartPicker.vue) and
+            // v-calendar (via FormControlDateRange.vue) are both now deferred
+            // at their call sites (defineAsyncComponent / no longer installed
+            // as an eager global plugin in Core.js), so this group is only
+            // downloaded once one of those lazy chunks actually loads — this
+            // is purely a caching optimization for chunks that already ship
+            // together, not an eager download.
             if (pkg === 'd3' || pkg.startsWith('d3-') || pkg === 'v-calendar' || pkg.startsWith('topojson')) {
               return 'vendor-charts'
```
pirhoo: important: not deferred yet: `@icij/murmur` in `vendor-ui` statically imports `d3-geo`/`d3-scale`, so the built `index.html` still modulepreloads `vendor-charts` (~400 kB) on every page and `v-calendar` rides along. Its own group delivers the saving Core.js describes:

```suggestion
            // Calendar, now only reachable through FormControlDateRange.vue's
            // own import since Core.js no longer installs the plugin globally.
            if (pkg === 'v-calendar') {
              return 'vendor-calendar'
            }

            // Charting/geo. Not deferred: @icij/murmur (vendor-ui) statically
            // imports d3-geo/d3-scale, so this group is preloaded everywhere.
            if (pkg === 'd3' || pkg.startsWith('d3-') || pkg.startsWith('topojson')) {
              return 'vendor-charts'
```

src/views/Search/SearchFilters.vue
```diff
@@ -31,8 +49,15 @@ const filters = computed(() => {
...
 
 const filtersTitles = computed(() => filters.value.map(filter => ({ filter, title: t(`filter.${filter.name}`) })))
-const fuse = computed(() => new Fuse(filtersTitles.value, { threshold: 0.1, shouldSort: false, keys: ['title'] }))
-const fuseFilters = computed(() => fuse.value.search(q.value).map(property('item.filter')))
+
+// Fuse.js is only needed once the user actually searches the filter list, so
+// defer loading its chunk until the first keystroke instead of importing it
+// eagerly for every search-page load.
+const Fuse = shallowRef(null)
+watch(q, () => import('fuse.js').then(module => (Fuse.value = module.default)), { once: true })
```
pirhoo: suggestion: with `{ once: true }` and no `.catch()`, one failed chunk load leaves `Fuse` null for good and the box silently lists every filter. Retrying on the next keystroke stays just as lazy:

```suggestion
watch(q, () => {
  if (!Fuse.value) {
    import('fuse.js').then(module => (Fuse.value = module.default))
  }
})
```

src/composables/useBreakpoints.js
```diff
@@ -20,6 +20,10 @@ const BREAKPOINT_VARIABLES = {
 
 // Function to get the CSS variable value
 const getCssBreakpointValue = (variable) => {
+  // ponytail: guards a real race, not a hypothetical one — this module is reached via
+  // lazy-loaded route chunks, and vitest can tear down jsdom (removing getComputedStyle)
+  // before such a pending import settles, throwing an unhandled ReferenceError in CI.
+  if (typeof document === 'undefined' || typeof getComputedStyle !== 'function') return 0
```
pirhoo: suggestion: the `ponytail:` marker leaked into the shipped comment, and there is no `curly` rule in `eslint.config.js` to catch the braceless `if`:

```suggestion
  // Guards a real race: this module is reached via lazy-loaded route chunks,
  // and vitest can tear down jsdom (removing getComputedStyle) before such a
  // pending import settles, throwing an unhandled ReferenceError in CI.
  if (typeof document === 'undefined' || typeof getComputedStyle !== 'function') {
    return 0
  }
```

pirhoo (commented):
Nice sweep on the mechanical side: all 90 `lodash/<fn>` deep paths resolve, the `orderBy` renames are consistent, and every d3 submodule export used exists at the installed versions. The notes below are all on the behavioural changes.

pirhoo (changes requested):
Requesting changes for the advanced-search modal regression flagged inline: with the new `v-if` gate the modal mounts already visible, so it never prefills or focuses on a first open and submitting overwrites the active query. The `vendor-charts` preload note is worth settling too, since it is the saving this PR advertises.

caro3801:
Thanks for your feedbacks

**Advanced search modal** – fixed in 375d24834. Added `{ immediate: true }` to `whenever(isVisible, handleOpen)` so mounting already-visible (which is how `SearchToolbar` renders it after the first open) triggers the same prefill/focus/formKey-bump path as a later open.

**`vendor-charts` preload** – only partially addressed (ffa78727a): `v-calendar` now gets its own `vendor-calendar` chunk instead of riding along in `vendor-charts`, and the comment now says plainly that `vendor-charts` is still preloaded everywhere.

The preload itself isn't fixable from this repo: the app's own code no longer has any bare `d3` imports (all deep `d3-*` submodule imports already), so the remaining `d3-geo`/`d3-scale` pull-in is `@icij/murmur`'s barrel statically re-exporting components this app doesn't use. We looked at a subpath-exports fix in murmur itself but decided not to pursue it for now, so this is a known, tracked, upstream-owned limitation rather than something to work around in `vite.config.js` here. 

Happy to open a murmur issue to track it if that's useful, but I'd rather not add a resolver/alias hack on the datashare-client side for a problem that belongs in murmur's build.

pirhoo (approved):
OK perfect thanks! Then go ahead with the merge! The sooner the better considering the size of the PR.

PR #481, opened 2026-08-11
spike: replace elasticsearch-browser with a thin axios client

caro3801:
Stacked on #479.

## What

While investigating bundle bloat on #479, `elasticsearch-browser` (1.6 MB, the single biggest item in the whole app bundle per the original measurement) turned out to be replaceable: this app's real dependency surface on it is just 4 primitives — a constructor, `get`, `search`, `count` — plus `transport.request` for the `_async_search` endpoints the classic client doesn't wrap. Everything else that looks like "the ES client API" (`getDocumentWithoutContent`, `searchFilter`, `rootSearch`, `estimateDownloadSize`, `countDocuments`, etc.) is Datashare's own `datasharePlugin` business logic in `elasticsearch.js`, unrelated to the package itself.

It's also worth noting `elasticsearch-browser` is effectively abandoned: latest version is 16.7.1, published March 2020, no `module`/`exports`/`sideEffects` fields at all. The maintained successor, `@elastic/elasticsearch`, explicitly does not support browser environments (its own README says so) and depends on Node-only HTTP internals (`undici`, `hpagent`) — not a viable swap.

## The change

- `src/api/elasticsearchClient.js` (new, ~150 lines): a `Client`/`Transport` pair built on `axios` (already a dependency) implementing `get`, `getSource`, `search`, `count`, plus the same `plugins` constructor hook the old client exposed — so `datasharePlugin`/`csrfPlugin` in `elasticsearch.js` keep patching `Client.prototype`/`Transport.prototype` completely unchanged.
- `src/api/elasticsearch.js`: only the import and the final `new Client(...)` call change.
- `elasticsearch-browser` moves from `dependencies` to `devDependencies` — its only remaining consumer is `tests/unit/specs/utils/esConnectionHelper.js` (an integration-test harness needing the real client's indices admin API), which never ships in the production bundle.
- Retry parity with `elasticsearch-browser`'s old transport-layer behavior: it silently retried every request up to 3 times on connection-level failures (never on an actual HTTP error response). The thin client's first version made exactly one attempt — a real regression, since fixed. `Transport`/`Client` now retry up to `maxRetries` (default 3, configurable) on network-level failures only, stop immediately on any HTTP error response or once `.abort()` is called, and reuse a single `AbortSignal` across all retries.
- Array query params now serialize as repeated `key=val` (`paramsSerializer: { indexes: null }`) instead of axios's default `key[]=val`, which ES rejects with "unrecognized parameter".

## Verification

- `yarn lint`: clean.
- Real `yarn build`, same machine, before/after:

  | | before | after |
  |---|---|---|
  | `vendor-search` chunk | 608.11 kB / 102.31 kB gzip | **27.13 kB / 7.56 kB gzip** |

  -95.5% raw / -92.6% gzip on that chunk.

- **Tested against a real ES cluster, not just mocks.** A live ES 8.19.8 is reachable in this dev environment at `http://elasticsearch:9200`. Ran every spec file that uses `esConnectionHelper` (the real-cluster integration harness), not just the ES-specific ones:
  - `elasticsearch.spec.js` + `elasticsearch.asyncSearch.spec.js`: pass (CSRF header injection, abort-signal forwarding, async_search transport calls, query building, error passthrough).
  - 8 more real-ES-backed files (`PathTree`, `DocumentContent`, `DocumentViewTabsEntities`, `WidgetFieldFacets`, `WidgetDocuments`, `WidgetDiskUsage`, `DocumentTranslation`, `ProjectDeletionModal`, `SearchBar`, `DocumentGlobalSearchTerms`, `DocumentThread`, `FilterType*`): 111/116 pass live against the real cluster. The 5 non-passes are a single pre-existing, confirmed-environmental flake (`window is not defined` post-teardown race in an unrelated component, `ParentOverflowEntries.vue`) — re-ran that file alone and got 5/5.
  - New: a live-ES round-trip test for `submitAsyncSearch`/`getAsyncSearch`/`deleteAsyncSearch` through the real `Transport`/axios path — the exact wire-format-mismatch bug class mocks can't catch.
  - New: `tests/unit/specs/api/elasticsearchClient.spec.js` — dedicated `Transport`/`Client` unit tests mocking only `axios`: host normalization, request shape, the array-param serializer fix, error passthrough, abort, and the full retry matrix (retry-until-success, give-up-after-default, custom `maxRetries`, no-retry-on-HTTP-error, abort-stops-retrying).
  - New: a regression test confirming a real (unmocked) 400 backend error propagates unchanged through `handleSearchError` — thrown and emitted on the event bus as the same object, not re-wrapped or swallowed.

## Bugs this coverage caught

Writing the tests above surfaced two real bugs in the first version of this client, both fixed here:
- **Missing `Client.getSource`** (`GET /:index/_source/:id`) — used by `DocumentTranslation.vue`/`documentDownload.js`, would have broken at runtime.
- **Array query params rejected by ES** — axios's default `_source[]=a&_source[]=b` format, fixed as noted above.

### DIFF
--- a/src/api/elasticsearchClient.js
+++ b/src/api/elasticsearchClient.js
@@ -64,7 +64,7 @@ async function requestWithRetries(config, attemptsLeft) {
  * patches (`Transport.prototype.request`) to inject a header on every call.
  */
 export class Transport {
-  constructor({ host, requestTimeout, maxRetries = DEFAULT_MAX_RETRIES } = {}) {
+  constructor({ host, requestTimeout = 30000, maxRetries = DEFAULT_MAX_RETRIES } = {}) {
     this.baseURL = /^https?:\/\//.test(host) ? host : `${window.location.protocol}//${host}`
     this.requestTimeout = requestTimeout
     this.maxRetries = maxRetries
--- a/src/api/elasticsearchClient.js
+++ b/src/api/elasticsearchClient.js
@@ -8,10 +8,9 @@ import axios from 'axios'
  * multi-transport support is used, so none of it needs to ship to the browser.
  *
  * Implements just enough of the classic client's shape (`Client.prototype`,
- * `components.Transport.prototype`, a `plugins` constructor option) that
- * `datasharePlugin`/`csrfPlugin` in elasticsearch.js keep working completely
- * unchanged — only this file and elasticsearch.js's final `new es.Client(...)`
- * call differ from the elasticsearch-browser-based version.
+ * `components.Transport.prototype`) that `datasharePlugin`/`csrfPlugin` in
+ * elasticsearch.js keep working unchanged, applied once at module scope
+ * there rather than per-instance (nothing here clones the prototype).
  */
 
 /**
@@ -106,9 +105,8 @@ export class Transport {
  * the async_search endpoints the plain client doesn't wrap.
  */
 export class Client {
-  constructor({ host, requestTimeout, maxRetries, plugins = [] } = {}) {
+  constructor({ host, requestTimeout, maxRetries } = {}) {
     this.transport = new Transport({ host, requestTimeout, maxRetries })
-    plugins.forEach(plugin => plugin(Client, {}, { Transport }))
   }
 
   get({ index, id, ...params }) {
--- a/src/api/elasticsearchClient.js
+++ b/src/api/elasticsearchClient.js
@@ -1,0 +1,99 @@
+import axios from 'axios'
+
+/**
+ * Minimal stand-in for elasticsearch-browser's `Client`/`Transport` pair,
+ * built on axios (already a dependency) instead of a bundled HTTP/transport
+ * stack. Datashare only ever talks to one known ES endpoint over plain HTTP —
+ * none of elasticsearch-browser's connection pooling, node sniffing, or
+ * multi-transport support is used, so none of it needs to ship to the browser.
+ *
+ * Implements just enough of the classic client's shape (`Client.prototype`,
+ * `components.Transport.prototype`, a `plugins` constructor option) that
+ * `datasharePlugin`/`csrfPlugin` in elasticsearch.js keep working completely
+ * unchanged — only this file and elasticsearch.js's final `new es.Client(...)`
+ * call differ from the elasticsearch-browser-based version.
+ */
+
+/**
+ * Drops undefined values so axios' query-string serializer omits the key
+ * entirely rather than serializing it as an empty string.
+ */
+function compactQuery(params = {}) {
+  return Object.fromEntries(Object.entries(params).filter(([, value]) => value !== undefined))
+}
+
+/**
+ * Issues raw HTTP requests to the ES endpoint. The one seam `csrfPlugin`
+ * patches (`Transport.prototype.request`) to inject a header on every call.
+ */
+export class Transport {
+  constructor({ host, requestTimeout } = {}) {
+    this.baseURL = /^https?:\/\//.test(host) ? host : `${window.location.protocol}//${host}`
+    this.requestTimeout = requestTimeout
+  }
+
+  /**
+   * @param {Object} params
+   * @param {string} [params.method='GET']
+   * @param {string} params.path - ES REST path, e.g. '/my-index/_search'
+   * @param {Object} [params.query] - Query-string parameters
+   * @param {Object} [params.body] - JSON request body
+   * @param {Object} [params.headers]
+   * @returns {Promise & { abort: Function }} Resolves with the response body;
+   *   `.abort()` cancels the in-flight request (mirrors the old client's
+   *   abortable-promise return value, which callers already rely on).
+   */
+  request({ method = 'GET', path, query, body, headers } = {}) {
+    const controller = new AbortController()
+    const promise = axios({
+      method,
+      baseURL: this.baseURL,
+      url: path,
+      params: compactQuery(query),
+      data: body,
+      headers,
+      timeout: this.requestTimeout,
+      signal: controller.signal
+    }).then(response => response.data)
+    promise.abort = () => controller.abort()
+    return promise
+  }
+}
+
+/**
+ * The only 4 primitives Datashare's own `datasharePlugin` actually builds on:
+ * a constructor, `get`, `search`, `count`, and `this.transport.request` for
+ * the async_search endpoints the plain client doesn't wrap.
+ */
+export class Client {
+  constructor({ host, requestTimeout, plugins = [] } = {}) {
+    this.transport = new Transport({ host, requestTimeout })
+    plugins.forEach(plugin => plugin(Client, {}, { Transport }))
+  }
+
+  get({ index, id, ...params }) {
+    return this.transport.request({
+      method: 'GET',
+      path: `/${index}/_doc/${encodeURIComponent(id)}`,
+      query: params
+    })
+  }
+
+  search({ index, body, ...params }) {
+    return this.transport.request({
+      method: 'POST',
+      path: `/${index}/_search`,
+      query: params,
+      body
+    })
+  }
+
+  count({ index, body, ...params }) {
+    return this.transport.request({
+      method: 'POST',
+      path: `/${index}/_count`,
+      query: params,
+      body
+    })
+  }
+}
--- a/src/api/elasticsearchClient.js
+++ b/src/api/elasticsearchClient.js
@@ -15,11 +15,13 @@ import axios from 'axios'
  */
 
 /**
- * Drops undefined values so axios' query-string serializer omits the key
- * entirely rather than serializing it as an empty string.
+ * Joins array values into comma-separated lists (`_source=a,b`): ES keeps
+ * only the last value of a repeated query param. Axios already drops
+ * undefined/null values from the query string.
  */
 function compactQuery(params = {}) {
-  return Object.fromEntries(Object.entries(params).filter(([, value]) => value !== undefined))
+  const joinArrays = ([key, value]) => [key, Array.isArray(value) ? value.join(',') : value]
+  return Object.fromEntries(Object.entries(params).map(joinArrays))
 }
 
 /**
@@ -87,10 +89,6 @@ export class Transport {
       baseURL: this.baseURL,
       url: path,
       params: compactQuery(query),
-      // ES rejects axios' default `key[]=val` array format ("unrecognized
-      // parameter: [_source[]]"); `indexes: null` serializes repeated params
-      // as `key=val1&key=val2` instead, which ES's _source/fields params expect.
-      paramsSerializer: { indexes: null },
       data: body,
       headers,
       timeout: this.requestTimeout,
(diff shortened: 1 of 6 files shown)


src/api/elasticsearchClient.js
```diff
@@ -0,0 +1,149 @@
...
+ * call differ from the elasticsearch-browser-based version.
+ */
+
+/**
+ * Drops undefined values so axios' query-string serializer omits the key
+ * entirely rather than serializing it as an empty string.
+ */
+function compactQuery(params = {}) {
+  return Object.fromEntries(Object.entries(params).filter(([, value]) => value !== undefined))
+}
```
pirhoo, line 24: crucial: looks like list params hit a wire-format trap: ES keeps only the last value of a repeated query param, so `_source=title&_source=path` silently returns just `path` (verified against a live 8.x cluster; elasticsearch-browser comma-joined these). `searchRootDocuments` passes `_source` as an array via `getDocumentsByIds`, so embedded-document rows lose title, contentType and contentLength today. Comma-joining here fixes it, axios already drops undefined on its own, and the `paramsSerializer` option plus its comment in `request()` can then go since no arrays reach the serializer anymore:

```suggestion
/**
 * Joins array values into comma-separated lists (`_source=a,b`): ES keeps only
 * the last value of a repeated query param. Undefined values need no special
 * handling, axios already omits them (and null) from the query string.
 */
function compactQuery(params = {}) {
  const joinArrays = ([key, value]) => [key, Array.isArray(value) ? value.join(',') : value]
  return Object.fromEntries(Object.entries(params).map(joinArrays))
}
```

pirhoo (commented):
Deep pass on the thin-client swap, focused on wire-format and behavior parity with elasticsearch-browser. Inline: 1 crucial (list params), 2 important (timeout retries and the test pinning the wrong format), plus smaller notes. Findings that sit outside the diff:

- suggestion: `vite.config.js` still lists `elasticsearch-browser` in the `vendor-search` manualChunks under the comment calling it the biggest dependency in the app; nothing under `src/` imports it anymore, so the entry is dead and the comment now misleads bundle profiling.
- suggestion: `elasticsearch.js` keeps a byte-identical `compactQuery` twin for the async-search wrappers; once the transport handles list and undefined params, the helper and its two call-site wraps can go.
- question: with AxiosError shapes, an ES-originated 401 now matches `handleHttpError` in App.vue and pops the persistent logged-out toast, which the old string-body errors never triggered. Intended upgrade or side effect?
- hint: `getDocument` wraps `return this.get(...)` in a synchronous try/catch, so a rejected promise never reaches `handleSearchError`; pre-existing and untouched by this PR, just noting it while in the area.
- remark: the live-cluster specs are a great catch-net, and currently the only coverage exercising `datasharePlugin` on the new transport, since `esConnectionHelper` still builds its shared client from elasticsearch-browser.

src/api/elasticsearchClient.js
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Whether a failed request is worth retrying: only connection-level failures
+ * (timeout, DNS, dropped connection - axios sets no `.response` for these),
+ * never an actual HTTP error response like a 400 or 500, which elasticsearch
+ * itself answered and won't answer differently on replay.
+ */
+function isRetryable(error) {
+  return !error?.response
+}
```
pirhoo, line 42: important: axios timeouts reject with code `ECONNABORTED` (or `ETIMEDOUT` with `clarifyTimeoutError`) and no `.response`, so they get retried here, while the old transport zeroed its retries on timeout and only ever replayed connection faults. With the 60s `requestTimeout`, one hung search now means up to 4 minutes of sequential attempts, and a timed-out `submitAsyncSearch` re-POSTs and can strand stored async searches on the cluster:

```suggestion
/**
 * Whether a failed request is worth retrying: only connection-level failures
 * (DNS, dropped connection - axios sets no `.response` for these), never a
 * timeout (the old client zeroed retries on timeout too), nor an actual
 * HTTP error response like a 400 or 500, which elasticsearch itself answered
 * and won't answer differently on replay.
 */
function isRetryable(error) {
  const timedOut = error?.code === 'ECONNABORTED' || error?.code === 'ETIMEDOUT'
  return !error?.response && !timedOut
}
```

tests/unit/specs/api/elasticsearchClient.spec.js
```diff
@@ -0,0 +1,191 @@
...
+      const transport = new Transport({ host: 'http://elasticsearch:9200' })
+      await transport.request({ path: '/my-index/_search', query: { _source: ['title', 'path'] } })
+
+      const { params, paramsSerializer } = axios.mock.calls[0][0]
+      expect(params).toEqual({ _source: ['title', 'path'] })
+      // This is the actual bug that shipped: without `indexes: null`, axios'
+      // default serializer produces `_source[]=title&_source[]=path`, which
+      // ES rejects as an "unrecognized parameter".
+      expect(paramsSerializer).toEqual({ indexes: null })
+    })
```
pirhoo: important: this pins down the format that loses fields: with last-wins semantics on repeated params, the repeated-key form is the bug rather than the fix, so the assertion will need to flip alongside the transport change:

```suggestion
    it('serializes array query params as comma-separated lists', async () => {
      const transport = new Transport({ host: 'http://elasticsearch:9200' })
      await transport.request({ path: '/my-index/_search', query: { _source: ['title', 'path'] } })

      // ES keeps only the last value of a repeated query param, so list
      // params go over the wire comma-joined, like the old client sent them.
      expect(axios.mock.calls[0][0].params).toEqual({ _source: 'title,path' })
    })
```

src/api/elasticsearchClient.js
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * The only 4 primitives Datashare's own `datasharePlugin` actually builds on:
+ * a constructor, `get`, `search`, `count`, and `this.transport.request` for
+ * the async_search endpoints the plain client doesn't wrap.
+ */
+export class Client {
+  constructor({ host, requestTimeout, maxRetries, plugins = [] } = {}) {
+    this.transport = new Transport({ host, requestTimeout, maxRetries })
+    plugins.forEach(plugin => plugin(Client, {}, { Transport }))
```
pirhoo: suggestion: every construction with plugins re-patches the shared prototypes, where elasticsearch-browser cloned the prototype per instance, so a second `new Client({ plugins })` (Vite HMR re-evaluating elasticsearch.js, a future second endpoint) stacks csrf wrappers on all instances, plugin-free ones included. Could we drop the `plugins` option and apply them once at module scope in elasticsearch.js, right above the singleton: `datasharePlugin(Client)` then `csrfPlugin(Client, {}, { Transport })`? Both plugin bodies stay byte-identical.

src/api/elasticsearchClient.js
```diff
@@ -0,0 +1,149 @@
...
+    return requestWithRetries(config, attemptsLeft - 1)
+  }
+}
+
+/**
+ * Issues raw HTTP requests to the ES endpoint. The one seam `csrfPlugin`
+ * patches (`Transport.prototype.request`) to inject a header on every call.
+ */
+export class Transport {
+  constructor({ host, requestTimeout, maxRetries = DEFAULT_MAX_RETRIES } = {}) {
```
pirhoo: suggestion: when `requestTimeout` is omitted axios gets `timeout: undefined` and never times out, where the old client defaulted to 30s; a default here keeps future bare constructions from hanging forever.

```suggestion
  constructor({ host, requestTimeout = 30000, maxRetries = DEFAULT_MAX_RETRIES } = {}) {
```

src/api/elasticsearchClient.js
```diff
@@ -0,0 +1,149 @@
...
+  return !error?.response
+}
+
+/**
+ * Issues one attempt, retrying on connection-level failure until `attemptsLeft`
+ * is exhausted or the shared signal has been aborted.
+ */
+async function requestWithRetries(config, attemptsLeft) {
+  try {
+    const response = await axios(config)
```
pirhoo, line 72: question: going through the global axios means ES traffic now inherits apiInstance's setup: its XSRF defaults already write the same cookie into the same header csrfPlugin adds (on same-origin hosts), plus the `withJsonCharset` interceptor. Is the plugin kept for a cross-origin `VITE_ES_HOST` in dev, or could it go? Either way a one-line comment here would spare the next reader the double-take.

pirhoo (changes requested):

src/api/elasticsearchClient.js
caro3801: Verified against elasticsearch-browser's own param spec: `_source` is declared type `list`, and its serializer joins arrays with `.join(',')`, so the old client comma-joined these too. Fixed in 5606d38 by joining arrays in `compactQuery` and dropping `paramsSerializer`; test flipped to assert the comma-joined format.

src/api/elasticsearchClient.js
caro3801: Confirmed, axios sets `code: ECONNABORTED`/`ETIMEDOUT` with no `.response` on timeout. Fixed in ec2ab01: `isRetryable` now excludes those codes, added a regression test for it.

tests/unit/specs/api/elasticsearchClient.spec.js
caro3801: Flipped alongside the transport fix in 5606d38

src/api/elasticsearchClient.js
caro3801: Only one construction happens today, but the risk is real under Vite HMR re-evaluating elasticsearch.js. Moved both plugins to module scope in 9ad9f67 and dropped the `plugins` constructor option entirely, since nothing else needs it.

src/api/elasticsearchClient.js
caro3801: Added the default in 2ce9cc9 with a regression test.

src/api/elasticsearchClient.js
caro3801: apiInstance's `axios.defaults.xsrf*` is same-origin only, so csrfPlugin still covers a cross-origin `VITE_ES_HOST` in dev. Kept the plugin and added the comment in a3adfab.

caro3801:
Pushed 3 more commits following up on the review:

- `9825e8cd` refactor: dropped `elasticsearch.js`'s local `compactQuery` helper, redundant now that `Transport.request` already drops undefined query params for every call.
- `0bfcc2ad` feat: reinstated `elasticsearch-browser`'s error shape on failed requests - `.status`, `.displayName` (`NotFound`, `RequestTimeout`, `ConnectionFault`, ...), and a `.message` built from ES's error body (`root_cause`) when present. Nothing in the app reads this today, but it's there so nothing regresses for future callers that do. Aborted (cancelled) requests are deliberately left undecorated.
- `6a21902d` test: locked host-resolution parity for a bare hostname with no port (verified it resolves identically to elasticsearch-browser's `Host`).

26 tests in `elasticsearchClient.spec.js`, 57 across the ES suite, all green. Lint clean.

pirhoo (approved):
Perfecto, thanks!

PR #483, opened 2026-08-11
perf(router): lazy-view regression guard, beforeEnter audit, hover/focus preload

caro3801:
Stacked on #481. 

## What's in this PR

1. **Regression guard for the search route's lazy named views** — `tests/unit/specs/router/routes.spec.js` walks the router's `routes` tree and asserts `Search.vue`/`SearchFilters.vue`/`SearchSettings.vue` and the nested `DocumentView` stay dynamic `import()`s, so an eager import creeping back in fails CI instead of only showing up in a bundle analyzer. Verified it actually catches the regression (temporarily reverted one loader to a static import, confirmed the test fails, reverted back).

2. **Audit of `beforeEnter` guards for redundant store writes** — verification only, no code change. Traced the full call graph: `updateFromRouteQuery` does fire twice on route entry (once via `prefillSearchStore`'s `beforeEnter`, once via the `onAfterRouteQueryUpdate` immediate handler), but it's harmless — `lastAppliedQuery` (what `sameAppliedQuery()` diffs against) is only ever written inside `doRefresh`, never touched by `updateFromRouteQuery`. So the duplication is a wasted, fully idempotent store reset/re-hydration, not a duplicate `refresh()`/network call. Findings recorded in `todo.md`.

3. **Preload nav target route chunks on hover/focus intent** — added a `preload(to)` helper in `AppSidebarSectionEntry.vue` (the shared component every primary nav link renders through — search, project list, task board, history, saved searches, etc.), wired to `@mouseenter`/`@focus` on both the main and optional action link. Calls `router.resolve(to).matched`'s component loader functions directly; no dedup bookkeeping needed since dynamic `import()` is idempotent via the browser's module cache. New spec (`AppSidebarSectionEntry.spec.js`, 4 tests) uses a throwaway route + spy loader to stay fast/deterministic rather than paying real chunk-transform cost.

## Testing

- `tests/unit/specs/router/routes.spec.js` — 2/2 pass
- `tests/unit/specs/components/AppSidebar/AppSidebarSectionEntry.spec.js` — 4/4 pass
- `tests/unit/specs/components/AppSidebar/AppSidebar.spec.js` (existing) — 2/2 pass, unaffected
- `yarn lint --fix` clean

### DIFF
--- a/src/components/AppSidebar/AppSidebarSectionEntry.vue
+++ b/src/components/AppSidebar/AppSidebarSectionEntry.vue
@@ -1,5 +1,6 @@
 <script setup>
 import { computed } from 'vue'
+import { useRouter } from 'vue-router'
 import { AppIcon } from '@icij/murmur'
 import IPhPlus from '~icons/ph/plus'
 
@@ -45,6 +46,22 @@ const classList = computed(() => {
     'app-sidebar-section-entry--exact-match': props.exactMatch
   }
 })
+
+const router = useRouter()
+
+// Preload a nav target's route chunk(s) on hover/focus intent, before the
+// user actually clicks. `router.resolve` + calling each matched record's
+// component loader triggers the same dynamic import() the router would run
+// on navigation; the browser's module cache makes repeat calls (re-hover,
+// actual navigation) a no-op, so no extra fetch or dedup bookkeeping needed.
+function preload(to) {
+  if (!to) return
+  for (const record of router.resolve(to).matched) {
+    for (const component of Object.values(record.components ?? {})) {
+      if (typeof component === 'function') component()
+    }
+  }
+}
 </script>
 
 <template>
@@ -55,6 +72,8 @@ const classList = computed(() => {
     <router-link
       :to="to"
       class="app-sidebar-section-entry__link text-truncate d-flex flex-grow-1"
+      @mouseenter="preload(to)"
+      @focus="preload(to)"
     >
       <app-icon
         class="app-sidebar-section-entry__link__icon me-2"
@@ -69,6 +88,8 @@ const classList = computed(() => {
       :to="actionTo"
       class="app-sidebar-section-entry__action ms-2 d-flex"
       :title="actionTitle"
+      @mouseenter="preload(actionTo)"
+      @focus="preload(actionTo)"
     >
       <app-icon
         class="app-sidebar-section-entry__action__icon"
--- a/src/components/AppSidebar/AppSidebarSectionEntry.vue
+++ b/src/components/AppSidebar/AppSidebarSectionEntry.vue
@@ -4,6 +4,8 @@ import { useRouter } from 'vue-router'
 import { AppIcon } from '@icij/murmur'
 import IPhPlus from '~icons/ph/plus'
 
+import { preload as vPreload } from '@/directives/preload'
+
 const props = defineProps({
   compact: {
     type: Boolean
@@ -49,19 +51,6 @@ const classList = computed(() => {
 
 const router = useRouter()
 
-// Preload a nav target's route chunk(s) on hover/focus intent, before the
-// user actually clicks. `router.resolve` + calling each matched record's
-// component loader triggers the same dynamic import() the router would run
-// on navigation; the browser's module cache makes repeat calls (re-hover,
-// actual navigation) a no-op, so no extra fetch or dedup bookkeeping needed.
-function preload(to) {
-  if (!to) return
-  for (const record of router.resolve(to).matched) {
-    for (const component of Object.values(record.components ?? {})) {
-      if (typeof component === 'function') component()
-    }
-  }
-}
 </script>
 
 <template>
@@ -70,10 +59,9 @@ function preload(to) {
     :class="classList"
   >
     <router-link
+      v-preload="{ to, router }"
       :to="to"
       class="app-sidebar-section-entry__link text-truncate d-flex flex-grow-1"
-      @mouseenter="preload(to)"
-      @focus="preload(to)"
     >
       <app-icon
         class="app-sidebar-section-entry__link__icon me-2"
@@ -84,12 +72,11 @@ function preload(to) {
     </router-link>
     <router-link
       v-if="actionTo"
+      v-preload="{ to: actionTo, router }"
       v-b-tooltip.body.right="{ delay: tooltipDelay }"
       :to="actionTo"
       class="app-sidebar-section-entry__action ms-2 d-flex"
       :title="actionTitle"
-      @mouseenter="preload(actionTo)"
-      @focus="preload(actionTo)"
     >
       <app-icon
         class="app-sidebar-section-entry__action__icon"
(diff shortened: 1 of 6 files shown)


src/components/AppSidebar/AppSidebarSectionEntry.vue
```diff
@@ -55,6 +72,8 @@ const classList = computed(() => {
     <router-link
       :to="to"
       class="app-sidebar-section-entry__link text-truncate d-flex flex-grow-1"
+      @mouseenter="preload(to)"
+      @focus="preload(to)"
```
pirhoo: important: this fires on the very first `mouseenter`/`focus` with nothing to cancel it, and hovering the search entry alone starts 260 chunk requests (~510 KiB raw on this branch's build, against a 6-chunk entry graph), so a sweep down the sidebar queues that per entry against the current view's own requests. Could we let the hover settle and drop it on leave, paired with the `preload` change above? The action link needs the same two handlers.

```suggestion
      @mouseenter="preload(to)"
      @mouseleave="cancelPreload"
      @focus="preload(to)"
      @blur="cancelPreload"
```

src/components/AppSidebar/AppSidebarSectionEntry.vue
```diff
@@ -45,6 +46,22 @@ const classList = computed(() => {
...
+// on navigation; the browser's module cache makes repeat calls (re-hover,
+// actual navigation) a no-op, so no extra fetch or dedup bookkeeping needed.
+function preload(to) {
+  if (!to) return
+  for (const record of router.resolve(to).matched) {
+    for (const component of Object.values(record.components ?? {})) {
+      if (typeof component === 'function') component()
+    }
+  }
+}
```
pirhoo: important: calling the loaders directly is fire-and-forget, so a failed chunk fetch (offline, stale hashes after a redeploy) lands as an uncaught rejection, and since failed module loads are not cached it re-fires on every re-hover, the opposite of what the comment promises. vue-router exports `loadRouteLocation` (4.6.4), which caches into the matched record and rejects cleanly, so it can carry the intent delay too. Needs `loadRouteLocation` on the vue-router import above.

```suggestion
// Preload a nav target's route chunk(s) on hover/focus intent, before the
// user actually clicks. Hover and focus are intent rather than a click, so
// let them settle first and drop the preload if the pointer moves on.
// `loadRouteLocation` runs the same dynamic import() the router would run
// on navigation and caches the result on the matched record, so repeat
// calls (re-hover, actual navigation) are a no-op.
const PRELOAD_DELAY = 80
let preloadTimer = null

function preload(to) {
  cancelPreload()
  preloadTimer = setTimeout(() => loadRouteLocation(router.resolve(to)).catch(() => {}), PRELOAD_DELAY)
}

function cancelPreload() {
  clearTimeout(preloadTimer)
}
```

tests/unit/specs/router/routes.spec.js
```diff
@@ -0,0 +1,39 @@
...
+    const search = findRoute(routes, 'search')
+    expect(isDynamicImport(search.components.default)).toBe(true)
+    expect(isDynamicImport(search.components.filters)).toBe(true)
+    expect(isDynamicImport(search.components.settings)).toBe(true)
+  })
+
+  it('lazy-loads the document view route', () => {
+    const document = findRoute(routes, 'document')
+    expect(isDynamicImport(document.component)).toBe(true)
+  })
```
pirhoo, line 38: important: the guard asserts loader shape rather than bundle placement, so it cannot catch the regression it advertises, and this case is already false while green: `Search.vue` statically imports `DocumentModal`, which statically imports `DocumentView`, so the compiled viewer ships inside the search chunk graph (its template is in `DocumentActionsGroup-*.js`, and no `DocumentView-*` chunk exists). Could we make it true by turning that `DocumentView` import in `DocumentModal.vue` into a `defineAsyncComponent(() => import(...))`, or drop the case?

src/components/AppSidebar/AppSidebarSectionEntry.vue
```diff
@@ -69,6 +88,8 @@ const classList = computed(() => {
       :to="actionTo"
       class="app-sidebar-section-entry__action ms-2 d-flex"
       :title="actionTitle"
+      @mouseenter="preload(actionTo)"
+      @focus="preload(actionTo)"
```
pirhoo: suggestion: in compact mode the section nav is `AppSidebarSectionToggler`, which renders a resolved `:href` rather than a `router-link` and so gets no preload, leaving the win dependent on which control the user hovers. Hoisting these bindings into one composable or a `v-preload` directive would cover the toggler too.

pirhoo (commented):
remark: nice catch on vitest's SSR transform rewriting `import(`, and on documenting why the matcher accepts both forms. That one would have been a confusing red. Notes below are on the preload itself and on what the new guard actually guarantees.

pirhoo (changes requested):
Requesting changes for the two inline notes: the preload has no intent delay and nothing cancels it (one hover starts 260 chunk requests, ~510 KiB), and the new document-route assertion is green while the viewer already ships inside the search chunk graph.

caro3801:
as suggested, the logic as been extracted the into `src/directives/preload.js` (`v-preload="{ to, router }"`): 80ms settle delay (intent), cancels on mouseleave/blur, applied to both router-links in `AppSidebarSectionEntry.vue` in commit [c142678](https://github.com/ICIJ/datashare-client/pull/483/commits/c1426785a7b7ee43dabce41ec19d126b89ef0301)

src/components/AppSidebar/AppSidebarSectionEntry.vue
caro3801: done in [c142678](https://github.com/ICIJ/datashare-client/pull/483/commits/c1426785a7b7ee43dabce41ec19d126b89ef0301)

src/components/AppSidebar/AppSidebarSectionEntry.vue
caro3801: extracted to directive . done in [c142678](https://github.com/ICIJ/datashare-client/pull/483/commits/c1426785a7b7ee43dabce41ec19d126b89ef0301)

tests/unit/specs/router/routes.spec.js
caro3801: `DocumentModal.vue` now loads `DocumentView` via `defineAsyncComponent(() => import(...))` so `DocumentView-*.js` is its own chunk, dynamically imported from` DocumentModal-*.js`, absent from `search-*.js/DocumentActionsGroup-*.js`. fixed here : https://github.com/ICIJ/datashare-client/pull/483/changes/c1426785a7b7ee43dabce41ec19d126b89ef0301#diff-d54e6c8e26ac75e62b6e12abb177e73dc6b30e16998c25b72d108f19f6d121f8R34

src/components/AppSidebar/AppSidebarSectionEntry.vue
caro3801: extracted to directive . done in [c142678](https://github.com/ICIJ/datashare-client/pull/483/commits/c1426785a7b7ee43dabce41ec19d126b89ef0301)

pirhoo (approved):
💯

PR #485, opened 2026-08-12
fix: remove router injection warning + a filter-panel flash bug

caro3801:
## PR description

**Fix eager `useSearchStore()` router injection warning + a filter-panel flash bug**

### Router injection warning
`search.js` called vue-router's `useRouter()` eagerly at store-setup time, before `Core.configure()` finishes and the router plugin installs, throwing `[Vue warn]: injection "Symbol(router)" not found.` on every boot. Fixed by deferring the `useRouter()` call into the one computed that actually uses it (`stringifyBaseRouteQuery`), instead of reordering boot (which was tried first and broke real navigation timing).

### Path filter flash on unrelated filter edits
Editing any filter (e.g. document type) visually closed and reopened the path filter's tree. Root cause: `setIndices()` reassigned `indices.value` to a new array on every route round-trip, even when the project selection hadn't changed. Consumers watching `searchStore.indices` by reference (`PathTree.vue`) treated this as a real change and reloaded, flashing the loading placeholder. Fixed by keeping the array reference stable on no-op calls.

Both fixes verified with regression tests (fail without the fix, pass with it) and, for the path-filter bug, live against a running dev server + backend.

### DIFF
--- a/src/store/filters/icons.js
+++ b/src/store/filters/icons.js
@@ -1,0 +1,39 @@
+import { markRaw } from 'vue'
+
+import IPhCirclesThreePlus from '~icons/ph/circles-three-plus'
+import IPhStar from '~icons/ph/star'
+import IPhHash from '~icons/ph/hash'
+import IPhUsers from '~icons/ph/users'
+import IPhTreeStructure from '~icons/ph/tree-structure'
+import IPhFile from '~icons/ph/file'
+import IPhFiles from '~icons/ph/files'
+import IPhCalendarBlank from '~icons/ph/calendar-blank'
+import IPhGlobe from '~icons/ph/globe'
+import IPhUserSquare from '~icons/ph/user-square'
+import IPhBuildings from '~icons/ph/buildings'
+import IPhMapPin from '~icons/ph/map-pin'
+import IPhEnvelope from '~icons/ph/envelope'
+import IPhPaperclip from '~icons/ph/paperclip'
+import IPhCalendarPlus from '~icons/ph/calendar-plus'
+
+import { CONTENT_TYPE_CATEGORY_FILTER_NAME } from '@/store/filters/FilterContentTypeCategory'
+
+// Icons for the built-in filters, keyed by filter `name`. Kept out of
+// `store/filters/index.js` so they don't ship in the eager core chunk.
+export default {
+  project: markRaw(IPhCirclesThreePlus),
+  starred: markRaw(IPhStar),
+  tags: markRaw(IPhHash),
+  recommendedBy: markRaw(IPhUsers),
+  path: markRaw(IPhTreeStructure),
+  contentType: markRaw(IPhFile),
+  [CONTENT_TYPE_CATEGORY_FILTER_NAME]: markRaw(IPhFiles),
+  creationDate: markRaw(IPhCalendarBlank),
+  language: markRaw(IPhGlobe),
+  namedEntityPerson: markRaw(IPhUserSquare),
+  namedEntityOrganization: markRaw(IPhBuildings),
+  namedEntityLocation: markRaw(IPhMapPin),
+  namedEntityEmail: markRaw(IPhEnvelope),
+  extractionLevel: markRaw(IPhPaperclip),
+  indexingDate: markRaw(IPhCalendarPlus)
+}
--- a/src/store/filters/icons.js
+++ b/src/store/filters/icons.js
@@ -20,7 +20,10 @@ import { CONTENT_TYPE_CATEGORY_FILTER_NAME } from '@/store/filters/FilterContent
 
 // Icons for the built-in filters, keyed by filter `name`. Kept out of
 // `store/filters/index.js` so they don't ship in the eager core chunk.
-export default {
+// Object.create(null): consumers index this map with user-typed field
+// names (e.g. SearchParameterFilter.vue), and a plain {} would resolve
+// prototype keys like `toString`/`constructor` instead of undefined.
+export default Object.assign(Object.create(null), {
   project: markRaw(IPhCirclesThreePlus),
   starred: markRaw(IPhStar),
   tags: markRaw(IPhHash),
@@ -36,4 +39,4 @@ export default {
   namedEntityEmail: markRaw(IPhEnvelope),
   extractionLevel: markRaw(IPhPaperclip),
   indexingDate: markRaw(IPhCalendarPlus)
-}
+})
--- a/src/store/modules/search.js
+++ b/src/store/modules/search.js
@@ -15,7 +15,6 @@ import toString from 'lodash/toString'
 import uniq from 'lodash/uniq'
 import lucene from 'lucene'
 import { ref, computed, toRaw } from 'vue'
-import { useRouter } from 'vue-router'
 
 import EsDocList from '@/api/resources/EsDocList'
 import { runAsyncSearch } from '@/api/asyncSearch'
@@ -150,20 +149,6 @@ export const useSearchStore = defineSuffixedStore('search', () => {
       fields: fields.value,
       operator: searchOperator.value
     }
-  })
-
-  const stringifyBaseRouteQuery = computed(() => {
-    const name = 'search'
-    const query = toBaseRouteQuery.value
-    // Router is resolved lazily here (rather than eagerly at store-setup
-    // time) because this store is instantiated during Core.configure(),
-    // before the router plugin is installed on the app — calling
-    // useRouter() at setup time throws a "Symbol(router) not found"
-    // injection warning on every boot. By the time anything actually reads
-    // this computed, the app (and router) are fully mounted.
-    const router = useRouter()
-    const { href = null } = router?.resolve({ name, query }) ?? {}
-    return href
   })
 
   const retrieveQueryTerms = computed(() => {
@@ -1113,7 +1098,6 @@ export const useSearchStore = defineSuffixedStore('search', () => {
     toRouteQuery,
     toRouteQueryWithStamp,
     toSearchParams,
-    stringifyBaseRouteQuery,
     retrieveQueryTerms,
     retrieveContentQueryTerms,
     page,
--- a/src/store/modules/search.js
+++ b/src/store/modules/search.js
@@ -26,6 +26,17 @@ import { defineSuffixedStore } from '@/store/defineSuffixedStore'
 import { SEARCH_OPERATORS } from '@/enums/searchOperators'
 import settings from '@/utils/settings'
 
+/**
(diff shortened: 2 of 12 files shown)


src/store/filters/icons.js
```diff
@@ -0,0 +1,39 @@
...
+  [CONTENT_TYPE_CATEGORY_FILTER_NAME]: markRaw(IPhFiles),
+  creationDate: markRaw(IPhCalendarBlank),
+  language: markRaw(IPhGlobe),
+  namedEntityPerson: markRaw(IPhUserSquare),
+  namedEntityOrganization: markRaw(IPhBuildings),
+  namedEntityLocation: markRaw(IPhMapPin),
+  namedEntityEmail: markRaw(IPhEnvelope),
+  extractionLevel: markRaw(IPhPaperclip),
+  indexingDate: markRaw(IPhCalendarPlus)
+}
```
pirhoo: important: the map is a plain object literal, so `builtinFilterIcons[field.value]` in `SearchParameterFilter.vue` resolves inherited keys for anything a user types in the search bar. `toString:foo` renders the chip as the literal text `[object Undefined]` instead of the magnifying glass, and `constructor:` / `__proto__:` give `Object` / `Object.prototype`. Could we drop the prototype once here so all four consumers are covered?

```suggestion
export default Object.assign(Object.create(null), {
  project: markRaw(IPhCirclesThreePlus),
  starred: markRaw(IPhStar),
  tags: markRaw(IPhHash),
  recommendedBy: markRaw(IPhUsers),
  path: markRaw(IPhTreeStructure),
  contentType: markRaw(IPhFile),
  [CONTENT_TYPE_CATEGORY_FILTER_NAME]: markRaw(IPhFiles),
  creationDate: markRaw(IPhCalendarBlank),
  language: markRaw(IPhGlobe),
  namedEntityPerson: markRaw(IPhUserSquare),
  namedEntityOrganization: markRaw(IPhBuildings),
  namedEntityLocation: markRaw(IPhMapPin),
  namedEntityEmail: markRaw(IPhEnvelope),
  extractionLevel: markRaw(IPhPaperclip),
  indexingDate: markRaw(IPhCalendarPlus)
})
```

src/store/modules/search.js
```diff
@@ -360,7 +366,7 @@ export const useSearchStore = defineSuffixedStore('search', () => {
    * @param {string} value - The index to set for the search.
    */
   function setIndex(value) {
-    indices.value = [value]
+    setIndices([value])
   }
```
pirhoo, line 367: important: routing through `setIndices` picks up its `compact()`, which silently changed `setIndex`'s contract: `''` now yields `[]` instead of `['']`, and `'a,b'` splits into two indices. `Core.js` calls `setIndex(this.getDefaultProject())`, and `getDefaultProject()` returns `''` for a user with no projects, so `indices` stays empty, which flips `allProjectsSelected` in `useSearchFilter.js` to true and makes `searchStore.index` `undefined` instead of `''`. Since `setIndex` is genuinely on the route-query path (`withRouteQuery('index', setIndex)`), it still needs the guard, just not the normalisation:

```suggestion
  function setIndex(value) {
    if (!isEqual(indices.value, [value])) {
      indices.value = [value]
    }
  }
```

src/store/modules/search.js
```diff
@@ -158,6 +157,13 @@ export const useSearchStore = defineSuffixedStore('search', () => {
...
+    // Router is resolved lazily here (rather than eagerly at store-setup
+    // time) because this store is instantiated during Core.configure(),
+    // before the router plugin is installed on the app — calling
+    // useRouter() at setup time throws a "Symbol(router) not found"
+    // injection warning on every boot. By the time anything actually reads
+    // this computed, the app (and router) are fully mounted.
+    const router = useRouter()
     const { href = null } = router?.resolve({ name, query }) ?? {}
     return href
   })
```
pirhoo: important: `useRouter()` inside a computed getter makes the result depend on where the computed is first read. Outside a setup or render context `inject()` returns undefined and the getter caches `href = null` until `toBaseRouteQuery` changes, so nothing recovers it. It works today only because `TaskBatchSearchForm.vue` reads `uri` during render before `createBatchSearch()` reads it from a click handler. Since that component is the only consumer and `toBaseRouteQuery` is already exposed, could we drop this getter and build `uri` there with a component-scoped `useRouter()`?

src/store/modules/search.js
```diff
@@ -43,7 +43,6 @@ export const useSearchStore = defineSuffixedStore('search', () => {
   const lastAppliedQuery = ref({})
 
   const appStore = useAppStore()
-  const router = useRouter()
   const searchBreadcrumbStore = useSearchBreadcrumbStore()
 
   const index = computed({
```
pirhoo, line 58: suggestion: the writable `index` setter just below still assigns `indices.value` directly, so it bypasses the new reference guard. Nothing writes it today, but delegating the setter to `setIndex(value)` keeps the two write paths from drifting.

tests/unit/specs/core/routerInjectionWarning.spec.js
```diff
@@ -0,0 +1,39 @@
+import { createCore } from '@/core'
+import { apiInstance as api } from '@/api/apiInstance'
+
+vi.mock('@/api/apiInstance', {
+  apiInstance: {
+    getUser: vi.fn(),
+    getSettings: vi.fn(),
+    getProject: vi.fn()
+  }
+})
```
pirhoo: suggestion: `vi.mock` takes a factory as its second argument, so this object literal is ignored and the whole `@/api/apiInstance` module gets automocked instead. The test still passes because automock hands back `vi.fn()`s, but the literal reads as if it were the contract. Same pattern in `main.spec.js` if you want to fix both:

```suggestion
vi.mock('@/api/apiInstance', () => ({
  apiInstance: {
    getUser: vi.fn(),
    getSettings: vi.fn(),
    getProject: vi.fn()
  }
}))
```

tests/unit/specs/store/filters/FilterContentTypeCategory.spec.js
```diff
@@ -12,8 +12,7 @@ import { findBoolShould, findTermsClause } from '~tests/unit/specs/utils/esQuery
 describe('FilterContentTypeCategory.js', () => {
   describe('breadcrumb icon', () => {
     it('uses the files icon so the breadcrumb chip reads as file types', () => {
-      const def = filtersDefs.find(d => d.options.name === 'contentTypeCategory')
-      expect(def.options.icon).toBe(IPhFiles)
+      expect(builtinFilterIcons[CONTENT_TYPE_CATEGORY_FILTER_NAME]).toBe(IPhFiles)
     })
```
pirhoo: suggestion: the rewritten assertion only proves the map has an entry, so deleting the `?? builtinFilterIcons[...]` fallback from `FilterModalTitle.vue` or `FilterType.vue` keeps this green while the breadcrumb chip loses its icon. Could we assert through one consumer instead, for instance mounting `FilterModalTitle` with the content-type-category filter and checking the rendered `app-icon` name?

src/store/modules/search.js
```diff
@@ -374,7 +380,17 @@ export const useSearchStore = defineSuffixedStore('search', () => {
...
+    // change. updateFromRouteQuery() calls this on every route round-trip
+    // (every filter edit re-applies `index`/`indices` from the URL), so an
+    // unconditional reassignment churns `indices.value`'s identity even
+    // when the project selection is unchanged. Consumers that watch
+    // `searchStore.indices` by reference (e.g. PathTree.vue, via
+    // `projects = computed(() => searchStore.indices)`) then see a false
+    // "changed" signal and reload/flash their content for no reason.
+    if (!isEqual(indices.value, cleaned)) {
+      indices.value = cleaned
+    }
```
pirhoo: remark: nice root-cause fix. Guarding the assignment rather than patching `PathTree.vue` means every `searchStore.indices` watcher stops seeing the phantom change, and the comment explaining the route-round-trip churn is exactly the context the next reader needs.

pirhoo (changes requested):
Requesting changes for the three `important:` comments inline. The rest are yours to take or leave.

src/store/filters/icons.js
caro3801: Done in 794524b9a. Wrapped the map in `Object.assign(Object.create(null), {...})` so indexing it with a user-typed field name (`toString`, `constructor`, etc.) can no longer resolve a prototype member instead of falling through to the `??` fallback.

src/store/modules/search.js
caro3801: Done in 46b751db0. `setIndex` now assigns `indices.value` directly with an isEqual guard instead of delegating to `setIndices`, whose `compact()` was turning `setIndex('')` into `[]` instead of `['']` and breaking the no-projects case in `Core.js`. Later folded into the shared `assignIfChanged` helper in b51569dd5.

src/store/modules/search.js
caro3801: Done in 68caad4dd. Dropped `stringifyBaseRouteQuery` from the store. `TaskBatchSearchForm.vue` was its only consumer and already holds its own `router` instance, so `uri` is now built there from `formSearchStore.toBaseRouteQuery` and a component-scoped `router.resolve()`.

src/store/modules/search.js
caro3801: Done in a3a270e05. The `index` computed setter now delegates to `setIndex` so both write paths share the same guard.

tests/unit/specs/core/routerInjectionWarning.spec.js
caro3801: Done in 2d615943f. Changed `vi.mock('@/api/apiInstance', {...})` to `vi.mock('@/api/apiInstance', () => ({...}))` so the mock factory actually runs instead of being silently ignored in favor of automock. Same fix applied to `main.spec.js`, which had the identical pattern.

tests/unit/specs/store/filters/FilterContentTypeCategory.spec.js
caro3801: Done in cf03eb520. Replaced the direct map assertion with a mount of `FilterModalTitle`, asserting the rendered `app-icon`'s `name` prop. Deleting a consumer's `?? builtinFilterIcons[...]` fallback now fails this test.

src/store/modules/search.js
caro3801: No action needed, this was a remark. Kept the same guard-then-assign shape; it is now shared between `setIndex` and `setIndices` via `assignIfChanged` (b51569dd5) per the follow-up hint from the review notes.

pirhoo (approved):
LGTM!

PR #487, opened 2026-08-13
fix(popover): open popovers on click and make them keyboard-safe

pirhoo:
Popovers used to open on hover and focus, which made them fire accidentally and left them impossible to dismiss with the keyboard. This switches every popover in the application to a click trigger and adds the keyboard handling bootstrap-vue-next does not provide.

- fix: open popovers on click and close them on click outside, through a single global `BPopover` default in `Core.useBootstrapVue()`
- fix: close the popover on Escape, wired on `document` since the content is teleported to `body`
- fix: hand focus back to the opener when Escape closes a popover, instead of stranding it on `body`
- fix: make the `DisplayEmail` target reachable by keyboard with `role`, `tabindex`, `aria-expanded` and an Enter/Space toggle
- refactor: drop the `@focus="blur"` workarounds in the document download and share entries, now that focus no longer opens a popover
- test: cover Escape-to-close, focus restoration, the click default and the `DisplayEmail` keyboard path

### DIFF
--- a/src/components/AppPopover/AppPopover.vue
+++ b/src/components/AppPopover/AppPopover.vue
@@ -1,5 +1,5 @@
 <script setup>
-import { computed, onBeforeUnmount, onMounted } from 'vue'
+import { computed, onBeforeUnmount, onMounted, ref, watch } from 'vue'
 
 import AppPopoverHeader from './AppPopoverHeader.vue'
 
@@ -25,12 +25,30 @@ const show = () => (modelValue.value = true)
 const hide = () => (modelValue.value = false)
 const toggle = () => (modelValue.value = !modelValue.value)
 
+// The element focused when the popover opened, so Escape can hand focus back.
+const elementFocusedBeforeOpening = ref(null)
+
+watch(modelValue, (isVisible) => {
+  if (isVisible) {
+    elementFocusedBeforeOpening.value = document.activeElement
+  }
+})
+
+// The content is teleported into document.body, so focus may sit inside it when
+// Escape is pressed. Closing would then strand focus on <body>, since the
+// library does no focus management of its own.
+const restoreFocus = () => {
+  elementFocusedBeforeOpening.value?.focus?.()
+  elementFocusedBeforeOpening.value = null
+}
+
 // The popover content is teleported to document.body, so a template-level
 // @keydown here would never catch the key. bootstrap-vue-next's floating UI
 // wires no keyboard handling at all, so Escape-to-close is added by hand.
 const handleKeydown = (event) => {
   if (event.key === 'Escape' && visible.value) {
     hide()
+    restoreFocus()
   }
 }
 
--- a/src/components/AppPopover/AppPopover.vue
+++ b/src/components/AppPopover/AppPopover.vue
@@ -1,5 +1,6 @@
 <script setup>
-import { computed, onBeforeUnmount, onMounted, ref, watch } from 'vue'
+import { computed, ref, useTemplateRef, watch } from 'vue'
+import { useEventListener } from '@vueuse/core'
 
 import AppPopoverHeader from './AppPopoverHeader.vue'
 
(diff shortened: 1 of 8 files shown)


mvanzalu (approved):

src/components/AppPopover/AppPopover.vue
```diff
@@ -24,6 +24,41 @@ const visible = computed(() => !!modelValue.value)
...
+
+// The popover content is teleported to document.body, so a template-level
+// @keydown here would never catch the key. bootstrap-vue-next's floating UI
+// wires no keyboard handling at all, so Escape-to-close is added by hand.
+const handleKeydown = (event) => {
+  if (event.key === 'Escape' && visible.value) {
+    hide()
+    restoreFocus()
+  }
+}
```
caro3801, line 76: **crucial:** A few things compound here.
This Escape handler is registered on `document`, so when a popover is teleported inside `DocumentModal.vue` (see `DocumentSharePopover.vue`'s `teleportTo` computed), pressing Escape to close just the popover also reaches `bootstrap-vue-next`'s own element-scoped Escape handler on the modal and closes it too.

Separately, `restoreFocus()` only runs from this Escape path, so the other two ways a popover closes - clicking outside (live now that the trigger is click) and the header's close button (`@hide="hide"`) - leave focus stranded inside the removed teleported content.

And this watcher only fires on `false -> true`, so a popover mounted already-open (as the component's own new spec does) never captures an opener at all.

Could we stop the keydown from bubbling, drive focus-restore off any close (not just Escape), and capture the opener immediately too?

```suggestion
const hide = () => (modelValue.value = false)
const toggle = () => (modelValue.value = !modelValue.value)

// The element focused when the popover opened, so closing can hand focus back
// regardless of how it was closed (Escape, click-outside, or the close button).
const elementFocusedBeforeOpening = ref(null)

watch(
  modelValue,
  (isVisible) => {
    if (isVisible) {
      elementFocusedBeforeOpening.value = document.activeElement
    } else {
      elementFocusedBeforeOpening.value?.focus?.()
      elementFocusedBeforeOpening.value = null
    }
  },
  { immediate: true }
)

// The popover content is teleported to document.body, so a template-level
// @keydown here would never catch the key. bootstrap-vue-next's floating UI
// wires no keyboard handling at all, so Escape-to-close is added by hand.
const handleKeydown = (event) => {
  if (event.key === 'Escape' && visible.value) {
    event.stopPropagation()
    hide()
  }
}
```

(No `restoreFocus()` call left in `handleKeydown` on purpose - the watcher's `else` branch now runs it for every close path once `modelValue` flips to `false`, so it still fires on Escape, just via `hide()` instead of a direct call.)

src/components/AppPopover/AppPopover.vue
```diff
@@ -24,6 +24,41 @@ const visible = computed(() => !!modelValue.value)
...
+
+onMounted(() => {
+  document.addEventListener('keydown', handleKeydown)
+})
+
+onBeforeUnmount(() => {
+  document.removeEventListener('keydown', handleKeydown)
+})
```
caro3801: **hint:** This repo already depends on `@vueuse/core` and uses `onKeyStroke` elsewhere for the same self-cleaning-listener pattern - it would also collapse the per-instance `document` listener into a single reactive binding, which matters once several `DocumentSharePopover`/`DocumentDownloadPopover` instances stay mounted on a search-results page.

Worth swapping in a follow-up (needs `onKeyStroke` added to the `@vueuse/core` import on line 3)?

```suggestion
onKeyStroke('Escape', handleKeydown)
```

src/components/Display/DisplayEmail.vue
```diff
@@ -19,6 +19,12 @@ const props = defineProps({
 })
 const { t } = useI18n()
 
+// The target is a span or a strong rather than a button, so Enter and Space do
+// not raise a click of their own. Popovers open on click only, which would
+// leave this one unreachable by keyboard without an explicit toggle.
+const isPopoverVisible = ref(false)
+const togglePopover = () => (isPopoverVisible.value = !isPopoverVisible.value)
```
caro3801: **important:** This popover hand-rolls its own open state (`isPopoverVisible`/`togglePopover`) on a raw `<b-popover>` instead of using `<app-popover>`, so it doesn't inherit the Escape-close/focus-restore this PR just added elsewhere. There's no way to dismiss it via keyboard once opened.

It also double-fires: a screen reader that synthesizes a click for Enter/Space activation on a `role="button"` span would trigger both the manual keydown handler and `b-popover`'s own native click listener (on via Core.js's global `BPopover: { click: true }`), toggling the popover open then immediately closed.

Two possible options: either switch to `<app-popover>` so it removes the need for this local state entirely or extract behavior in composable so it's reusable

If moved to app-popover: `AppPopover` doesn't currently forward `placement`/`boundary-padding` to the underlying `b-popover`, so those two props would need to be added there first, or the migration will drop them silently.

src/components/AppPopover/AppPopover.vue
caro3801: **hint:** This Escape/focus-restore fix lives only in `AppPopover`, so raw `<b-popover>` consumers elsewhere (`I18nLocaleDropdown.vue`, `KeyboardShortcutsPopover.vue`) stay keyboard-undismissable and focus-unsafe.

Could this move to a small shared composable (e.g. `usePopoverEscape`) those consumers can also adopt, so every popover in the application have the same behavior?

caro3801 (changes requested):
Great additions on the popover accessibility work, the Escape-close and focus-restore pattern in `AppPopover.vue` is a clean fix. 

Before merging, though, there are still blocking issues: Escape currently also closes the enclosing modal, and focus doesn't get restored on click-outside or the close button, only on Escape. Everything else is smaller stuff, either non-blocking or good to tackle as a follow-up.

src/components/AppPopover/AppPopover.vue
pirhoo: Applied in 3d5e6a324, with one adjustment on each half.

Escape: a bubble-phase `stopPropagation` on `document` cannot protect the modal, because BModal's Escape handler is scoped to the modal element (`onKeyStroke('Escape', ..., { target: element })`) and element listeners run before anything document-level in bubble order. The listener is now registered on `document` in capture phase, which runs first, so the popover consumes the key and a second Escape closes the modal.

Focus: the opener is captured immediately and restore is driven by the watcher on every close as suggested, but it only fires when focus would otherwise be stranded on `<body>` or inside the teleported content. The click-outside close lands on window `click`, after mousedown has already focused the clicked control, so an unconditional restore would steal focus from e.g. an input the user just clicked into. Both behaviors are covered in the spec.

src/components/AppPopover/AppPopover.vue
pirhoo: Swapped in 3d5e6a324, via `useEventListener` rather than `onKeyStroke`: `OnKeyStrokeOptions` only exposes `eventName`/`target`/`passive`/`dedupe`, and the capture-phase registration is what keeps Escape from reaching a surrounding modal (see the other thread). The target is a getter returning `null` while closed, so mounted-but-closed popovers hold no document listener at all.

src/components/Display/DisplayEmail.vue
pirhoo: Moved to `app-popover` in 98bcdbbd5, dropping the local state. `placement`/`boundary-padding` needed no new props: they are declared `BPopover` props, so Vue's fallthrough attrs on `app-popover`'s root deliver them.

The double-fire is fixed one level lower: the target is now a real `<button>`, so Enter, Space and assistive-tech activation all raise a single native click handled once by the click trigger, and the manual keydown handlers are deleted, leaving no second path to race.

pirhoo:
Thanks for the quick review, I've made the changes :)

src/components/AppPopover/AppPopover.vue
caro3801: TIL thanks!

caro3801 (approved):

PR #488, opened 2026-08-13
feat(project): show project lists in a popover

### LINKED ISSUE #2323: feat: projects popover
Follow design here: https://github.com/ICIJ/datashare/issues/2320

pirhoo:
Stacked on #487, review that one first.

Long project lists used to be rendered inline and overflow their container. They now collapse into a thumbnail stack that opens a popover with the full list.

Closes icij/datashare#2323

- feat: add `ProjectsButton`, a thumbnail stack plus a count that opens the project list in a popover
- feat: use it in the task lists, the task board and the batch search cards
- refactor: extract `ProjectThumbnailStack`, shared with `ProjectDropdownSelectorButtonContent`
- refactor: replace the `DisplayProjectList` alias with the shared `toProjectList` helper in `utils/projects`
- fix: reuse the translated projects count key instead of a second hardcoded label
- fix: add a gap between the search bar addons and the submit button
- test: cover `ProjectsButton` and `ProjectThumbnailStack`, with stories for both

## Preview

Capture d’écran 2026-08-13 à 21 42 30
Capture d’écran 2026-08-13 à 21 42 34


mvanzalu (approved):

src/views/Task/TaskBoard/TaskBoardLatest.vue
caro3801, line 62: hint: Looks like `[item.args?.defaultProject]` becomes `[undefined]` whenever `defaultProject` is missing, and that flows into `ProjectButton`'s `props.project.name ?? props.project`, which throws on `undefined` and crashes the row. Could we filter it the way `resolveProject` (already used elsewhere in this PR) would?

```suggestion
      return [item.args?.defaultProject].filter(Boolean)
```
Low probability considering our batch search format but easy consolidation

src/components/Project/ProjectsButton.vue
```diff
@@ -0,0 +1,83 @@
...
+      </b-button>
+    </template>
+    <div class="projects-button__list d-flex flex-wrap gap-2">
+      <project-button
+        v-for="(project, index) in projectList"
+        :key="index"
+        :project="project"
+      />
+    </div>
+    <template #close>
```
caro3801: suggestion: In `AppPopover`/`AppPopoverHeader` could we add a `hideClose` prop to alongside the existing `hideHeader`, to prevent this trick?

src/components/Project/ProjectThumbnailStack.vue
```diff
@@ -0,0 +1,88 @@
...
+  return props.overflow && projectList.value.length > props.max
+})
+</script>
+
+<template>
+  <span class="project-thumbnail-stack">
+    <project-thumbnail
+      v-for="(project, index) in visibleProjects"
+      :key="index"
+      :project="resolveProject(project, core)"
```
caro3801: suggestion: Looks like `resolveProject(project, core)` runs inline in this `v-for` expression, so it re-resolves every visible project on every re-render, where `ProjectLabel.vue`'s identical call is memoized in a `computed()` 

Could this follow the same pattern?

src/utils/projects.js
```diff
@@ -0,0 +1,21 @@
+import isObject from 'lodash/isObject'
+
+/**
+ * Normalize a project prop, which may be a bare name, a project object, or a
+ * list of either, into a list.
+ */
+export function toProjectList(projects) {
+  return [].concat(projects ?? [])
```
caro3801: crucial: Looks like `projects ?? []` only catches `null`/`undefined` — an empty string isn't nullish, so `toProjectList('')` returns `['']` and renders a phantom project button wherever a task's `defaultProject` is blank 

In addition `toProjectList` hand-rolls what the codebase elsewhere gets from lodash's `castArray`, reusing it (with a falsy guard) would fix the empty-string gap above and match the existing convention in one line.

* src/components/Widget/WidgetDocumentsByCreationDateByPath.vue:111 : return `castArray(this.project)` for instance

```suggestion
  return projects ? castArray(projects) : []
```
with lodash castArray import

src/utils/projects.js
```diff
@@ -0,0 +1,21 @@
+import isObject from 'lodash/isObject'
+
```
caro3801, line 3:
```suggestion
import castArray from 'lodash/castArray'
```

src/components/Project/ProjectsButton.vue
```diff
@@ -0,0 +1,83 @@
...
+const projectList = computed(() => toProjectList(props.projects))
+
+const label = computed(() => {
+  return t('searchBarInputDropdownForProjects.projectsCount', projectList.value.length)
+})
+</script>
+
+<template>
+  <project-button
+    v-if="projectList.length === 1"
```
caro3801, line 33: suggestion: With only `v-if="=== 1"`/`v-else-if="> 1"` and no fallback, zero projects renders nothing here, unlike the deleted `DisplayProjectList` which always rendered a container (eventhough the translation key has been added).
Only `SettingsSnapshotsList.vue` added its own `-` placeholder around this component, so every other consumer (batch download/search lists, task board, entities list) goes silently blank instead. 
Could we add a shared empty-state branch here?

caro3801 (changes requested):
Nice feature built on the new popover.
Worth addressing the projectList empty string issue. 
I put few suggestions that are consolidating what already there. By convention we never needed it but at least the gap will be filled.
Thanks for adding all the unit test and updating the stories

src/views/Task/TaskBoard/TaskBoardLatest.vue
pirhoo: Good catch, and it does crash rather than just render oddly: `ProjectButton`'s `to` computed reads `props.project.name` (`ProjectButton.vue:37`), which throws on `undefined`.

I fixed it one level up instead, in `toProjectList` (105c65a76), so this call site needs no change and the sibling task lists get it too, `TaskDocumentsList.vue:96` and `TaskEntitiesList.vue:103` build their list the same way:

```js
return castArray(projects).filter(Boolean)
```

Covered by the new `tests/unit/specs/utils/projects.spec.js`.

src/components/Project/ProjectsButton.vue
pirhoo: Done in 6f21b5d0b: `hideClose` on both `AppPopover` and `AppPopoverHeader`, and the `<span />` trick is gone. The existing "no close button" test asserts on the rendered DOM, so it keeps covering this unchanged.

src/components/Project/ProjectThumbnailStack.vue
pirhoo: Done in 3513bdcb2, mapped inside `visibleProjects` so it matches `ProjectLabel.vue:41` and `resolveProject` no longer leaks into template scope.

src/utils/projects.js
pirhoo: Agreed on `castArray`, adopted in 105c65a76.

One tweak to the suggestion: `projects ? castArray(projects) : []` fixes the bare `''` case but still keeps holes *inside* a list, so the `[item.args?.defaultProject]` from `TaskBoardLatest` survives as `[undefined]` and crashes `ProjectButton`. Filtering covers both, and makes the nullish guard redundant since `castArray(undefined)` is `[undefined]`:

```js
return castArray(projects).filter(Boolean)
```

New spec covers `''`, `[undefined]` and `['foo', undefined, '']`.

src/utils/projects.js
pirhoo: Done in 105c65a76.

src/components/Project/ProjectsButton.vue
pirhoo: Done in 95f1e371b.

Went with the `-` placeholder rather than the `no projects` label, to match the convention already used in this same table (`SettingsSnapshotsList.vue:102` does it for dates). That let me drop the now-duplicated local guard, so the cell is just `<projects-button :projects="item.indices" />`.

caro3801 (approved):

PR #493, opened 2026-08-18
fix: prevent flaky-adjacent timeout in useSearchBreadcrumb spec

### LINKED ISSUE #2337: Flaky-adjacent: useSearchBreadcrumb test runs near the 10s Vitest timeout
## Description

`tests/unit/specs/composables/useSearchBreadcrumb.spec.js`'s test `"clears only the contentTypeCategory value when the category chip is removed"` runs close to Vitest's 10s default timeout (~9.4s measured), on a commit that predates and is unrelated to the locked-filters work (ICIJ/datashare#2219) that surfaced it. It passes today but is one slow CI runner away from a flaky failure.

Likely cause (unconfirmed): first-in-file router/async-component warm-up cost rather than an actual hang — no async work in the test's own code path appears slow enough to account for it on its own.

## Suggested fix

Either raise this test's (or file's) timeout explicitly above the default, or warm up the router/async components once in a `beforeAll` for the file so the cost isn't paid inside a single test's timeout window.

## Acceptance criteria

- [x] The test reliably completes well under whatever timeout it runs against, not within ~94% of the default.

caro3801:
Fixes icij/datashare#2337

`clearEntry`'s `refreshRoute()` navigates to the `search` route, which lazily imports the whole Search view subtree on first resolution - that cost was landing inside a single test (~9.4s, vs the 10s Vitest timeout). Added a `beforeAll` that resolves it once on a throwaway router/pinia, so subsequent `beforeEach` navigations hit a warm import cache.

Test time dropped from ~9.4s to ~15ms for the affected test.

### DIFF
--- a/tests/unit/specs/composables/useSearchBreadcrumb.spec.js
+++ b/tests/unit/specs/composables/useSearchBreadcrumb.spec.js
@@ -1,11 +1,24 @@
 import { mount } from '@vue/test-utils'
+import { createRouter, createWebHashHistory } from 'vue-router'
+import { createPinia, setActivePinia } from 'pinia'
 
 import CoreSetup from '~tests/unit/CoreSetup'
 import { useSearchBreadcrumb } from '@/composables/useSearchBreadcrumb'
 import { useSearchStore } from '@/store/modules'
+import { routes } from '@/router'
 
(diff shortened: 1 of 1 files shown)


pirhoo (approved):
Nice!

PR #494, opened 2026-08-19
feat: render the document text tab as markdown

pirhoo:
Renders the document Text tab as formatted markdown when the server has a structure artifact for the document, with a toolbar dropdown to go back to plain text. Local and global search both work on the markdown pages, marked the same way the plain text view marks them.

- feat: add a structure artifact probe composable, shared by the download popover and the Text tab
- feat: render one markdown structure page at a time, sanitized (no raw HTML, no remote images, hardened links)
- feat: move the text rendering choice into a toolbar dropdown, disabled for translations
- feat: search inside the markdown pages, with page-aware next/previous navigation
- feat: mark local and global search terms in the rendered page, folding case and diacritics like the backend does
- feat: report an expired session from the artifact requests, which otherwise degrade a failure to an empty result
- fix: render pipe tables that have no header row, and drop a table header that renders with no content
- fix: fall back to plain text when the structure artifact holds no markdown
- test: cover the utils, the composable and both components

> [!NOTE]  
> We agreed with @Soliine to move the switch to plan/formatted text to a dropdown menu, following the same logic as the one in the custom PDF viewer. This dropdown menu will hold other "text" option such as centering content, wrapping text, or copying content.

## Preview

Capture d’écran 2026-08-19 à 11 33 33

### DIFF
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -409,6 +409,18 @@ export class Api {
     return data
   }
 
+  /**
+   * Count a query's occurrences in a document's markdown structure pages,
+   * per page. Bypasses `sendAction` because it runs on every (throttled)
+   * keystroke: the caller degrades a failure to zero occurrences instead of
+   * toasting it.
+   */
+  async searchStructurePages(index, id, query, routing) {
+    const url = Api.getFullUrl(`/api/${index}/artifacts/structure/search/${id}`)
+    const { data } = await this.axios.request({ url, method: Method.GET, params: { query, routing } })
+    return data
+  }
+
   login(username, password) {
     return this.sendAction('/auth/login', {
       method: Method.POST,
--- a/src/api/index.js
+++ b/src/api/index.js
@@ -405,7 +405,7 @@ export class Api {
    */
   async getStructurePage(index, id, page, routing) {
     const url = Api.getFullUrl(`/api/${index}/artifacts/structure/${id}/${page}`)
-    const { data } = await this.axios.request({ url, method: Method.GET, params: { routing }, responseType: 'text' })
+    const { data } = await this.requestArtifact({ url, params: { routing }, responseType: 'text' })
     return data
   }
 
@@ -417,8 +417,25 @@ export class Api {
    */
   async searchStructurePages(index, id, query, routing) {
     const url = Api.getFullUrl(`/api/${index}/artifacts/structure/search/${id}`)
-    const { data } = await this.axios.request({ url, method: Method.GET, params: { query, routing } })
+    const { data } = await this.requestArtifact({ url, params: { query, routing } })
     return data
+  }
+
+  /**
+   * Request an artifact endpoint outside `sendAction`, while still reporting an
+   * expired session on the error bus. Without this, callers that degrade a
+   * failure to "no results" would swallow a 401 and never offer a re-login.
+   */
+  async requestArtifact(config) {
+    try {
+      return await this.axios.request({ method: Method.GET, ...config })
+    }
+    catch (error) {
+      if (error.response?.status === 401) {
+        this.eventBus?.emit('http::error', error)
+      }
+      throw error
+    }
   }
 
   login(username, password) {
--- a/src/components/Document/DocumentContent.vue
+++ b/src/components/Document/DocumentContent.vue
@@ -72,7 +72,7 @@ const markdownMatches = ref([])
 const markdownAppliedTerm = ref('')
 // Set at the end of `onMounted`, so the mode watcher can tell a real, later mode
 // flip apart from the manifest probe settling `isMarkdownMode` during the mount.
-const isMounted = ref(false)
+let isMounted = false
 
 const isTranslation = computed(() => {
   return !!props.targetLanguage && props.targetLanguage !== 'original'
@@ -240,7 +240,7 @@ watch(isMarkdownMode, async (markdown) => {
   await syncPagePosition(markdown)
   // `hasMarkdown` flips during `onMounted`'s manifest probe, so without this gate
   // that flip would duplicate the search the mount is about to issue itself.
-  if (isMounted.value && hasLocalSearchTerms.value) {
+  if (isMounted && hasLocalSearchTerms.value) {
     await retrieveOccurrencesAndUpdateContent()
   }
 })
@@ -254,7 +254,16 @@ watch(docId, async () => {
   markdownPage.value = 1
   markdownMatches.value = []
   markdownAppliedTerm.value = ''
-  await fetchManifest()
+  maxOffsetTranslations.value = {}
+  syncedPages.value = []
+  localSearchIndexes.value = []
+  localSearchOccurrences.value = 0
+  localSearchIndex.value = 0
+  // An occurrence fetch issued for the document being left would otherwise
+  // resolve into the state of the one arriving: no other watcher supersedes it
+  // when the search term itself is unchanged across the swap.
+  lastOccurrencesRetrieval++
+  await loadDocumentContent()
 })
 
 watch(contentPipeline, async () => {
@@ -266,19 +275,31 @@ onMounted(async () => {
   // `finally`, because a mount step rejecting must not leave the flag stuck
   // `false` and suppress every later mode-flip search for this instance.
   try {
-    await Promise.all([loadMaxOffset(), fetchManifest()])
-    await syncPages()
-    if (props.q) {
-      await retrieveOccurrencesAndUpdateContent()
-    }
-    else {
-      await activateContentSlice({ offset: 0 })
-    }
+    await loadDocumentContent()
   }
   finally {
-    isMounted.value = true
+    isMounted = true
   }
 })
+
+// Shared with the `docId` watcher: this sequence describes a document, not a
+// mount, and the host can swap the document without remounting.
+async function loadDocumentContent() {
+  await Promise.all([loadMaxOffset(), fetchManifest()])
+  await syncPages()
+  // Slices are keyed by offset alone, so a watcher woken by the reset above can
+  // have cached one sliced against the length and page map of the previous
+  // document. Only now are both of them known for this one.
+  Object.keys(contentSlices).forEach(key => delete contentSlices[key])
+  activeContentSliceOffset.value = 0
+  currentContentPage.value = ''
+  if (hasLocalSearchTerms.value) {
+    await retrieveOccurrencesAndUpdateContent()
+  }
+  else {
+    await activateContentSlice({ offset: 0 })
+  }
+}
 
 // A single-page artifact whose only page renders to nothing has no markdown
 // worth showing, so the tab goes back to plain text. With more pages, a blank
--- a/src/components/Document/DocumentContent.vue
+++ b/src/components/Document/DocumentContent.vue
@@ -209,8 +209,8 @@ watch(page, async () => {
   await activateContentSlice({ offset })
 })
 
-watch(isMarkdownMode, (markdown) => {
-  syncPagePosition(markdown)
+watch(isMarkdownMode, async (markdown) => {
+  await syncPagePosition(markdown)
 })
 
 watch(contentPipeline, async () => {
@@ -232,14 +232,25 @@ onMounted(async () => {
 // Both paginations describe the same physical pages when their counts match,
 // so the page number survives the toggle; anything else has no page
 // correspondence and goes back to the first page.
-function syncPagePosition(markdown) {
+async function syncPagePosition(markdown) {
+  // A translation forces text mode through its own contract (start at offset
+  // 0, handled by the targetLanguage watcher): its offsets describe another
+  // language than the one `syncedPages`/`offsets` were computed for, so this
+  // function must not touch the page position in that case.
+  if (isTranslation.value) {
+    return
+  }
   const aligned = !!syncedPages.value?.length && syncedPages.value.length === markdownPagesCount.value
   if (markdown) {
     markdownPage.value = aligned ? offsets.value.indexOf(activeContentSliceOffset.value) + 1 || 1 : 1
+    return
   }
-  else {
-    activeContentSliceOffset.value = aligned ? offsets.value[markdownPage.value - 1] ?? 0 : 0
-  }
+  // The restored page needs its text slice loaded, and going through
+  // `activateContentSlice` also registers this offset as the current
+  // activation, so the activation the mode flip triggered for the previous
+  // offset cannot clobber it once it resolves.
+  const offset = aligned ? offsets.value[markdownPage.value - 1] ?? 0 : 0
+  await activateContentSlice({ offset })
 }
 
 const loadMaxOffset = waitFor(async function (targetLanguage = props.targetLanguage) {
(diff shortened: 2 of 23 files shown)


src/api/index.js
```diff
@@ -405,10 +405,39 @@ export class Api {
...
+    const { data } = await this.requestArtifact({ url, params: { query, routing } })
+    return data
+  }
+
+  /**
+   * Request an artifact endpoint outside `sendAction`, while still reporting an
+   * expired session on the error bus. Without this, callers that degrade a
+   * failure to "no results" would swallow a 401 and never offer a re-login.
+   */
+  async requestArtifact(config) {
```
caro3801, line 429: important: Looks like a 401 here fires both the global "log back in" toast (via `http::error`) and
`useDocumentDownload.js`'s own `toastedPromise` error toast for the same failure, so a user whose
session expires mid-download sees two stacked, partly contradictory messages.

The session-expiry toast could take precedence and have `toastedPromise` skip its own
message for that case (e.g. check `error.response?.status !== 401` before showing the generic
toast), so only one message reaches the user?

src/api/index.js
```diff
@@ -405,10 +405,39 @@ export class Api {
...
+    try {
+      return await this.axios.request({ method: Method.GET, ...config })
```
caro3801, line 431: suggestion: `requestArtifact` reimplements, just for the 401 case, the same try/catch -> `eventBus.emit('http::error', ...)` -> rethrow shape that `sendAction` already centralizes for every other endpoint.

Could this fold into `sendAction` (or a shared helper both call) instead of living as a second, parallel error-reporting path, so a future change to session-expiry handling only needs to happen once?

src/components/Document/DocumentContent.vue
```diff
@@ -53,6 +57,38 @@ const searchStore = useSearchStore.inject()
...
+const isMarkdownEmpty = ref(false)
+const markdownPage = ref(1)
+const markdownMatches = ref([])
+// The term the occurrence counts were computed for. The marks a page renders
+// have to match what the backend counted, so this only moves when a search
+// actually ran, not on every keystroke.
+const markdownAppliedTerm = ref('')
+// Set at the end of `onMounted`, so the mode watcher can tell a real, later mode
+// flip apart from the manifest probe settling `isMarkdownMode` during the mount.
+const isMounted = ref(false)
```
caro3801: hint: `isMounted` is only ever set once in `onMounted`'s `finally` and read inside an async watch callback, never from the template, so it doesn't need to be reactive.

```suggestion
let isMounted = false
```

(and drop the `.value` at the two read/write sites, lines 243 and 279)

src/components/Document/DocumentContent.vue
```diff
@@ -173,26 +229,92 @@ watch(toRef(props, 'targetLanguage'), async (value) => {
...
+// document identity re-runs it and clears what belonged to the previous one.
+watch(docId, async () => {
+  preferMarkdown.value = true
+  isMarkdownEmpty.value = false
+  markdownPage.value = 1
+  markdownMatches.value = []
+  markdownAppliedTerm.value = ''
+  await fetchManifest()
+})
+
```
caro3801, line 268: crucial: This watcher resets the markdown-specific refs on a document swap, but looks like it leaves the plain-text pagination state (`maxOffsetTranslations`, `syncedPages`, `contentSlices`, `activeContentSliceOffset`, `currentContentPage`) untouched. 

Since `DocumentContent` can be reused across documents without remounting, the previous document's page/offset state (and, further down, `syncPagePosition`'s alignment check against it) can leak into the new document's view. 

This also folds in the [comment below](https://github.com/ICIJ/datashare-client/pull/494/changes#r3813541102): bumping `lastOccurrencesRetrieval` here stops a stale occurrence response from the old document writing over the new one's state.

```suggestion
watch(docId, async () => {
  preferMarkdown.value = true
  isMarkdownEmpty.value = false
  markdownPage.value = 1
  markdownMatches.value = []
  markdownAppliedTerm.value = ''
  maxOffsetTranslations.value = {}
  syncedPages.value = []
  Object.keys(contentSlices).forEach(key => delete contentSlices[key])
  activeContentSliceOffset.value = 0
  currentContentPage.value = ''
  localSearchIndexes.value = []
  localSearchOccurrences.value = 0
  localSearchIndex.value = 0
  lastOccurrencesRetrieval++
  await Promise.all([loadMaxOffset(), fetchManifest()])
  await syncPages()
  if (hasLocalSearchTerms.value) {
    await retrieveOccurrencesAndUpdateContent()
  }
  else {
    await activateContentSlice({ offset: 0 })
  }
})
```

(`lastOccurrencesRetrieval` is declared with `let` further down the file, but since this callback only runs after setup finishes, the declaration is already in scope by the time it executes.) 

This duplicates most of `onMounted`'s body (lines 265-281) rather than sharing it, which is a fair follow-up cleanup, but keeping the two separate avoids reshaping `onMounted` under a crucial fix.

src/components/Document/DocumentContent.vue
```diff
@@ -307,37 +429,118 @@ async function retrieveOccurrencesAndUpdateContent() {
...
+  // A mode flip mid-flight can start the other path's retrieval before this
+  // one's response lands; this counter mirrors `lastContentSliceActivation`
+  // so only the newest retrieval is allowed to write its results, whichever
+  // response arrives last.
+  const retrieval = ++lastOccurrencesRetrieval
+  if (isMarkdownMode.value) {
+    return retrieveMarkdownOccurrences(retrieval)
+  }
+  return retrieveTextOccurrences(retrieval)
+}
```
caro3801, line 479: important: `lastOccurrencesRetrieval` only gets bumped from inside `retrieveTotalOccurrences`, and the `docId` watcher never calls it. If the search term is unchanged across a document swap (so its own watcher never re-fires), an occurrence fetch still in flight for the previous document can
resolve later and write into `localSearchOccurrences`/`localSearchIndexes`/`markdownMatches` for the new document.

Covered by the `loadDocumentContent()` suggestion in [previous comment](https://github.com/ICIJ/datashare-client/pull/494/changes#r3813527314), which bumps `lastOccurrencesRetrieval` on every document swap.

src/components/Document/DocumentContentMarkdown.vue
```diff
@@ -0,0 +1,319 @@
...
+  const locallyMarked = addSearchMarksClassInHtml(html, props.term)
+  // Global marks go on last so they nest inside the local ones, the order the
+  // `extracted-text` pipeline chain produces for the plain text view. Unlike that
+  // chain, a `regex` term is matched literally: this marker walks text nodes so it
+  // never marks inside an href, which a regex over rendered HTML would.
+  return props.globalSearchTerms.reduce((marked, { label }, index) => {
+    const style = `border-color: ${getTermIndexColor(index)}`
+    return addSearchMarksClassInHtml(marked, label, { className: 'global-search-term', style })
+  }, locallyMarked)
+})
```
caro3801, line 82: suggestion(perf): `markedHtml` calls `addSearchMarksClassInHtml` once for the local term and once per
`globalSearchTerms` entry via `reduce`, and each call does its own full `DOMParser().parseFromString` + tree walk + `innerHTML` serialization over the whole page. With
several active global search terms that's several full parse/walk/serialize passes on every recompute.

Would it be worth updating `addSearchMarksClassInHtml` to accept a list of terms and mark all of
them in one parse/walk/serialize pass, rather than reparsing per term?

src/components/Document/DocumentContentMarkdown.vue
```diff
@@ -0,0 +1,319 @@
...
+}
+
+// Plugins can transform the markdown body through the `markdown-text` category;
+// core registers nothing under it, so by default this resolves to the marked
+// HTML unchanged.
+async function cookHtml(html) {
+  cookedHtml.value = await pipelinesStore.applyPipelineChainByCategory('markdown-text')(html)
+  await nextTick()
+  activateMatch()
+}
```
caro3801, line 164: important: `cookHtml` (watched with `watch(markedHtml, cookHtml, { immediate: true })` further down) has no staleness guard, unlike the sibling `loadPage` right above it which already uses a
`lastPageLoad` counter for exactly this reason. If the pipeline call (or a future async step in it)
takes variable time, a slower call can resolve after a newer one and overwrite the page currently on screen.

The same token pattern as `loadPage` closes the gap:

```suggestion
let lastCook = 0

async function cookHtml(html) {
  const cook = ++lastCook
  const cooked = await pipelinesStore.applyPipelineChainByCategory('markdown-text')(html)
  if (cook !== lastCook) {
    return
  }
  cookedHtml.value = cooked
  await nextTick()
  activateMatch()
}
```

src/components/Document/DocumentContentMarkdown.vue
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    :deep(h6) {
+      font-size: 0.85rem;
+      color: var(--bs-secondary-color);
+    }
+
+    :deep(table) {
+      border-collapse: collapse;
+      margin-bottom: $spacer;
```
caro3801: suggestion: This heading/table/blockquote SCSS block is duplicated near-verbatim in
`DocumentViewerMarkdown.vue` (same selectors, margins, font sizes), and the two already drifted once in this PR's own history before being patched back in sync.

Could this move into a shared partial (e.g. `_markdown-body.scss`) that both components `@use`, so a future visual tweak can't apply to only one of them?

src/composables/useStructureArtifact.js
```diff
@@ -0,0 +1,49 @@
...
+      return
+    }
+    const fetch = ++lastFetch
+    const fetched = await api.getStructureManifest(index, id, routing)
+    // A probe for a document the user has already left can resolve after the
+    // one for the document on screen, so only the newest may write.
+    if (fetch === lastFetch) {
+      manifest.value = { ...fetched, documentId: id }
+    }
+  }
```
caro3801, line 44: suggestion: The "only the newest async response may write" counter (`let x = 0; const token = ++x;
...await...; if (token !== x) return`) shows up four times independently: here as `lastFetch`, and
as `lastOccurrencesRetrieval`/`lastContentSliceActivation` in `DocumentContent.vue` and
`lastPageLoad` in `DocumentContentMarkdown.vue`.

Using a shared helper could be help DRY (e.g. `useLatestOnly()` returning a `guard(promiseFn)`
wrapper), so a future fix to the race only needs applying once instead of by hand in four places?

src/composables/useStructureArtifact.js
```diff
@@ -0,0 +1,49 @@
...
+    return documentId === documentRef.value?.id && pages > 0 && formats.includes(MARKDOWN_FORMAT)
+  })
+
+  const pages = computed(() => {
+    return hasMarkdown.value ? manifest.value.pages : 0
+  })
+
+  let lastFetch = 0
+
+  async function fetchManifest() {
```
caro3801, line 32: hint: `fetchManifest` has no memoization by document id, and `useStructureArtifact` is instantiated separately in both `DocumentContent.vue` and `useDocumentDownload.js` for the same document, so opening a document and then its download popover probes the same manifest twice.

Worth caching the last manifest per document id (even a small module-level `Map`) so the second
instance can skip the round-trip?

caro3801 (changes requested):
Nice work getting the markdown tab and its pagination/search wiring this far, the inline comments made this an easy diff to follow, really well tested.

There are some changes to apply before merging. The one to prioritize before merge is [the watcher reset](https://github.com/ICIJ/datashare-client/pull/494/changes#r3813527314) (and its [close cousin](https://github.com/ICIJ/datashare-client/pull/494/changes#r3813541102)): the docId watcher resets the markdown state but not the text-pagination state, so a document swap without a remount can show the wrong page or stale search results.
The [double toast](https://github.com/ICIJ/datashare-client/pull/494/changes#r3813491545) and [`cookHtml` missing staleness](https://github.com/ICIJ/datashare-client/pull/494/changes#r3813562141) guard are worth a look too since they touch user-visible correctness, just narrower in scope.

The rest are good-to-have cleanups (shared counter helper, shared SCSS, fewer duplicate manifest fetches) that would keep this code easy to maintain but shouldn't hold up the PR.

src/components/Document/DocumentContent.vue
pirhoo: Fixed in fc645cd6c. The reset now clears the text-pagination state too and bumps `lastOccurrencesRetrieval`, and `onMounted` and the watcher share one `loadDocumentContent()`. Slices are cleared again after the manifest and max offset land, otherwise a watcher woken by the reset caches a slice sized against the previous document. Regression test: `should not carry the previous document page and slices over to the next one`.

src/components/Document/DocumentContent.vue
pirhoo: Fixed in fc645cd6c, via the `lastOccurrencesRetrieval++` in the reset as you suggested.

src/components/Document/DocumentContentMarkdown.vue
pirhoo: Fixed in 334fa6a13, with your `lastCook` pattern.

src/api/index.js
pirhoo: Fixed in 0761b7b0f. Dropped `toastedPromise` here and toast directly unless the status is 401, so the "log back in" toast stands alone.

src/components/Document/DocumentContentMarkdown.vue
pirhoo: Fixed in 5e2fd2f99. Moved to `src/utils/markdown.scss` as a `.markdown-body` class both views apply; the two style blocks are down to their own layout bits.

src/components/Document/DocumentContent.vue
pirhoo: Fixed in fc645cd6c.

src/api/index.js
pirhoo: Leaving as is: the two have opposite policies. `sendAction` reports every rejection, and the artifact endpoints exist precisely not to, since 404 is the common case. Folding them needs a status filter for one caller.

src/components/Document/DocumentContentMarkdown.vue
pirhoo: Leaving as is: the re-parse is what produces the nesting. After the local pass the `<mark>` is a real element, so the global pass wraps text nodes already inside it. One pass would have to resolve the overlaps by hand, for one artifact page and a handful of terms.

src/composables/useStructureArtifact.js
pirhoo: Leaving as is: a cached negative would keep the markdown tab hidden for a document whose artifact is created after the first probe. Not worth one saved round-trip without an invalidation story.

src/composables/useStructureArtifact.js
pirhoo: Leaving as is for now: the four sites differ in shape, one passes the token into two branch functions and one needs it after a `try/catch`, so a `guard(promiseFn)` wrapper does not fit. A `claim()` returning an `isLatest()` predicate would. Happy to do it in a follow-up if you want it.

caro3801 (approved):
lgtm 🥳

pirhoo:
Great review again, thanks!

src/composables/useStructureArtifact.js
caro3801: This is non blocking and no need for a follow up for now, thanks

PR #496, opened 2026-08-19
fix(search): submit the advanced search through the router

### LINKED ISSUE #2338: bug: query disappear when field is selected and next page is reached
**Describe the bug**
https://github.com/user-attachments/assets/81c56d41-02fb-4be9-b17b-951e42b10dce

**To Reproduce**
1. Go to [Search in LuxLeaks](https://datashare-demo.icij.org/#/?from=75&perPage=25&sort=_score&order=asc&searchOperator=OR&q=&indices=luxleaks)
2. Search for PDF as exact name or "Any of these words"
3. Tick Fields > Name
4. See results and click to the next page
5. See error: the query disappeared from the search bar and from the breadcrumb
6. Note: it doesn't happen when we use the field dropdown in the search bar. see video:
https://github.com/user-attachments/assets/d01acc80-1d07-4a34-b6d4-5b0f4b00f6e8

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser: Helium
 - Version: 21.16.0

lschim:
Fixes https://github.com/ICIJ/datashare/issues/2338

The advanced search modal ran its query straight from the store, leaving the URL on the previous query. Any later navigation that patches the current URL — pagination, sort, perPage — then re-applied that stale query over the advanced one, so the second page of results showed the previous search.

Forward the modal's submit to the search view, which updates the store and pushes a route the way the search bar already does. The URL stays the source of truth, the refresh guards run the search, and `from` resets to page one.

Fix made using Claude

### DIFF
--- a/src/views/Search/Search.vue
+++ b/src/views/Search/Search.vue
@@ -113,6 +113,20 @@ const page = useUrlPageFromWithStore({
 
 const documentViewFloatingId = provideDocumentViewFloatingId()
 
+/**
+ * Run an advanced search submitted from the toolbar's modal. Like the search
+ * bar, it only updates the store and pushes a route: the refresh guards below
+ * turn that navigation into the actual search. Pushing keeps the URL in sync
+ * with the store, without which the next route round-trip (pagination, sort,
+ * perPage) would re-apply the URL's previous query over this one.
+ */
+function handleAdvancedSearch({ query, field }) {
+  searchStore.setQuery(query)
+  searchStore.setField(field)
+  // Resets `from` and stamps the route, so an unchanged query still resubmits.
+  refreshRouteFromStart()
+}
+
 // Refresh route query when a filter changes (either their values or if they are excluded)
 watchFilters(refreshRouteFromStart)
 // Refresh route query when projects change
@@ -157,6 +171,7 @@ onConsumeNoRefresh({ immediate: route.name === 'search' })
           v-model:is-filters-closed="isFiltersClosed"
           :search-breadcrumb-counter="searchBreadcrumbCounter"
           :no-search-filters="!enoughFloatingSpace && !isSearchRoute"
+          @advanced-search="handleAdvancedSearch"
         />
         <search-breadcrumb
           v-model:visible="toggleSearchBreadcrumb"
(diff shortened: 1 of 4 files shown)


pirhoo (approved):
LGTM!

src/views/Search/Search.vue
```diff
@@ -113,6 +113,20 @@ const page = useUrlPageFromWithStore({
...
 const documentViewFloatingId = provideDocumentViewFloatingId()
 
+/**
+ * Run an advanced search submitted from the toolbar's modal. Like the search
+ * bar, it only updates the store and pushes a route: the refresh guards below
+ * turn that navigation into the actual search. Pushing keeps the URL in sync
+ * with the store, without which the next route round-trip (pagination, sort,
+ * perPage) would re-apply the URL's previous query over this one.
+ */
+function handleAdvancedSearch({ query, field }) {
```
caro3801, line 123: suggestion: `handleAdvancedSearch` shares the same behavior as [`SearchBar.vue`'s `submit()`](https://github.com/ICIJ/datashare-client/blob/main/src/components/Search/SearchBar/SearchBar.vue#L125-L137): commit the store then push a fresh route. 
Non blocking refactor suggestion : could we factorize this logic into the composable `refreshRouteFromStart()` or similar and use it in `SearchBar.vue` `submit()` too, so the two components can't drift apart?

caro3801 (approved):
Thanks for fixing this bug and adding a regression test. The suggestion is non blocking because it's still easy to read.

src/views/Search/Search.vue
lschim: If done in refreshRouteFromStart(), SearchBar would import useSearchFilter which is a big composable with watchers and SearchBar is reused at other places where there are no filters no? So, should I create a new Composable for the update of the store then pushing the new route ?
(I don't know if it is a real logic in Vue to think in term of cost of watchers, but I'm fairly certain that they are not free in term of performances)

caro3801: Yes good point on the watchers, I agree with you, a smaller composable by pulling refreshRoute/refreshRouteFromStart out of useSearchFilter would be a good fit instead of importing the everything from useSearchFilter

PR #497, opened 2026-08-20
fix(document): PDF viewer page tracking

### LINKED ISSUE #2342: bug: in Document > View, pagination bugs
**Describe the bug**
If I type N in the [TinyPagination](https://icij.github.io/datashare-client/?path=/docs/components-murmur-tinypagination--docs), sometimes the displayed number is N-1 and I am brought to the N-1th page and sometimes, it displays a random number of page (but it leads me to the right page).

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser: Helium
 - Version: 21.18.0

pirhoo:
Fixes the page indicator in the PDF viewer, which drifted on small windows and counted every page a scroll flew past. Closes ICIJ/datashare#2342.

- fix: track the page under a probe line below the toolbar, instead of any page merely touching the viewport (the `threshold` option had no effect, vueuse reports `isIntersecting`)
- fix: hold the indicator on the target page while a page pick or a search occurrence jump is scrolling, then realign, since lazily rendered pages resize mid-animation and the scroll lands short
- fix: only cycle search occurrences on Enter while the search field has focus, so Enter in the page field no longer does both

### DIFF
--- a/src/components/Document/DocumentViewer/DocumentViewerPdf.vue
+++ b/src/components/Document/DocumentViewer/DocumentViewerPdf.vue
@@ -36,6 +36,7 @@ const props = defineProps({
 
 const SCROLL_IDLE_DELAY = 200
 const SCROLL_KEYS = ['ArrowUp', 'ArrowDown', 'PageUp', 'PageDown', 'Home', 'End', ' ']
+const FORM_CONTROLS = 'input, textarea, select, [contenteditable]'
 
 const documentViewStore = useDocumentViewStore()
 const src = computed(() => (documentViewStore.embeddedPdf ? null : props.document.fullUrl))
@@ -115,9 +116,20 @@ function releasePendingPage() {
   currentPage.value = currentPageBelowToolbox() ?? currentPage.value
 }
 
+/**
+ * Whether a key press scrolls the document, as opposed to moving a caret inside a form control:
+ * a space typed in the search field must not count as the reader scrolling away.
+ *
+ * @param {KeyboardEvent} event - The key press to test.
+ * @returns {boolean}
+ */
+function scrollsDocument({ key, target }) {
+  return SCROLL_KEYS.includes(key) && !target?.closest?.(FORM_CONTROLS)
+}
+
 useEventListener(window, 'scroll', settleScroll, { capture: true, passive: true })
 useEventListener(window, ['wheel', 'touchmove', 'pointerdown'], releasePendingPage, { passive: true })
-useEventListener(window, 'keydown', ({ key }) => SCROLL_KEYS.includes(key) && releasePendingPage())
+useEventListener(window, 'keydown', event => scrollsDocument(event) && releasePendingPage())
 
 /**
  * Holds the page tracking until the scroll we are about to start has settled, so the indicator
@@ -202,8 +214,13 @@ watch(highlightIndex, () => {
   }
 })
 
-// A new document swaps every page out without them reporting they stopped showing.
-watch(pdf, () => pagesBelowToolbox.clear())
+// A new document swaps every page out without them reporting they stopped showing, and a hold left
+// over from the previous one would settle against pages that are not there anymore.
+watch(pdf, () => {
+  pagesBelowToolbox.clear()
+  pendingPage.value = null
+  realignPendingPage.value = false
+})
 
 watch(src, async () => {
   blurred.value ??= await isBlurred(props.document)
--- a/src/components/Document/DocumentViewer/DocumentViewerPdf.vue
+++ b/src/components/Document/DocumentViewer/DocumentViewerPdf.vue
@@ -1,7 +1,6 @@
 <script setup>
-import VueScrollTo from 'vue-scrollto'
 import { computed, ref, useTemplateRef, toRef, watch } from 'vue'
-import { refDebounced, whenever, useElementBounding } from '@vueuse/core'
+import { refDebounced, whenever, useDebounceFn, useElementBounding, useEventListener } from '@vueuse/core'
 import { supportsPDFs as embeddable } from 'pdfobject'
 import { useI18n } from 'vue-i18n'
 
@@ -14,7 +13,6 @@ import { SCALE_FIT, SCALE_WIDTH } from '@/enums/documentViewerPdf'
 import { useCompact } from '@/composables/useCompact'
 import { useDocumentPreview } from '@/composables/useDocumentPreview'
 import { useWait } from '@/composables/useWait'
-import { useScrollParent } from '@/composables/useScrollParent'
 import { usePDF } from '@/composables/usePDF'
 import { useDocumentViewStore } from '@/store/modules/documentView'
 import AppWait from '@/components/AppWait/AppWait'
@@ -36,8 +34,9 @@ const props = defineProps({
   }
 })
 
+const SCROLL_IDLE_DELAY = 200
+
 const documentViewStore = useDocumentViewStore()
-const scrollParent = useScrollParent()
 const src = computed(() => (documentViewStore.embeddedPdf ? null : props.document.fullUrl))
 const { pdf, numPages, sizes, findHighlights, loaderId: pdfLoaderId } = usePDF(src)
 const { waitFor, isLoading } = useWait()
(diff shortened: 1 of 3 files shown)


src/components/Document/DocumentViewer/DocumentViewerPdf.vue
```diff
@@ -105,6 +195,16 @@ whenever(highlightTextDebounced, waitFor(async (value) => {
...
 
+// Picking another occurrence makes the matching page scroll its highlight into view.
+watch(highlightIndex, () => {
+  if (highlightPage.value) {
+    holdPageTracking(highlightPage.value)
+  }
+})
+
+// A new document swaps every page out without them reporting they stopped showing.
+watch(pdf, () => pagesBelowToolbox.clear())
```
caro3801: crucial: Looks like `pendingPage`/`realignPendingPage` never get cleared when the  changes, only `pagesBelowToolbox` does.

Since this component can stay mounted across a document-to-document navigation (nothing keys it to the document identity), a hold started on the old document (e.g. via `scrollToPage`) can still be "pending" when the user jumps to the next document with the keyboard shortcut (`d+ArrowRight` / `Meta+ArrowRight`). 
`ArrowRight` isn't in `SCROLL_KEYS` and isn't a wheel/touch/pointerdown event either, so nothing releases the hold before the new document mounts.
When `settleScroll` later fires, it calls `scrollPageIntoView(pendingPage.value, 'instant')` against the *new* document's pages and unconditionally sets `currentPage.value = pendingPage.value` with no bounds check, so the pagination indicator can show a page number that does not exist in the new document.

Could we reset both refs in this same watcher, alongside `pagesBelowToolbox.clear()`?

```suggestion
watch(pdf, () => {
  pagesBelowToolbox.clear()
  pendingPage.value = null
  realignPendingPage.value = false
})
```

src/components/Document/DocumentViewer/DocumentViewerPdf.vue
```diff
@@ -58,29 +61,119 @@ const highlightTextDebounced = refDebounced(highlightText, 300)
...
+ * started by hand is never undone by the pending page settling behind it.
+ */
+function releasePendingPage() {
+  pendingPage.value = null
+  currentPage.value = currentPageBelowToolbox() ?? currentPage.value
+}
+
+useEventListener(window, 'scroll', settleScroll, { capture: true, passive: true })
+useEventListener(window, ['wheel', 'touchmove', 'pointerdown'], releasePendingPage, { passive: true })
+useEventListener(window, 'keydown', ({ key }) => SCROLL_KEYS.includes(key) && releasePendingPage())
```
caro3801: important: This `keydown` listener checks only the key name, not the event target, so it also fires while the user is typing in a text field rather than scrolling the document.

`DocumentLocalSearch`'s input and murmur's `PaginationTiny` page field are plain native inputs with
no `keydown` handling, so a space typed into a multi-word search query, or an arrow-key edit in the
page-number field, bubbles up to `window` and calls `releasePendingPage()` even though no document
scroll happened. That drops a page-tracking hold mid-settle, reintroducing the "counts every page a
scroll flew past" symptom this PR is meant to fix.

Could we ignore the event when it originates from a form control?

```suggestion
useEventListener(window, 'keydown', (event) => {
  if (!SCROLL_KEYS.includes(event.key) || event.target?.closest('input, textarea, select, [contenteditable]')) {
    return
  }
  releasePendingPage()
})
```

src/components/Document/DocumentViewer/DocumentViewerPdf.vue
```diff
@@ -105,6 +195,16 @@ whenever(highlightTextDebounced, waitFor(async (value) => {
   highlightIndex.value = highlightMatches.value.length ? 1 : 0
 }))
 
+// Picking another occurrence makes the matching page scroll its highlight into view.
+watch(highlightIndex, () => {
```
caro3801, line 211: suggestion: Cycling to a search occurrence doesn't update `currentPage` right away the way picking a page from the toolbar does.

`scrollToPage` sets `currentPage.value = value` immediately for instant feedback, but this `highlightIndex` watcher only arms `pendingPage`/`settleScroll` and leaves `currentPage` to catch up once the 200ms idle settle (or a stray wheel/keydown) fires. While that window lasts, the pagination widget shows the previous page number even though the viewer is already scrolling to the new one.


Could this call `currentPage.value = highlightPage.value` alongside `holdPageTracking`, the same way `scrollToPage` does, for a consistent instant update?

src/components/Document/DocumentViewer/DocumentViewerPdf/DocumentViewerPdfPage.vue
```diff
@@ -88,18 +95,16 @@ function applyHighlight() {
   const highlight = highlights[props.highlightIndex - 1]
   // Remove all classes "highlight--active" from the page and highlight the current one
```
caro3801, line 92: hint: This highlight-centering scroll still goes through `vue-scrollto` + `useScrollParent`, while the sibling page-jump scroll in `DocumentViewerPdf.vue` was migrated to native `scrollIntoView` plus the `--document-viewer-pdf-toolbox-height` scroll-margin.

Keeping both mechanisms around means any future change to the toolbox-offset handling has to be kept in sync by hand across two code paths. Not blocking for this PR, but worth a follow-up to move `applyHighlight` onto the same native `scrollIntoView` approach and let `vue-scrollto` / `useScrollParent` be dropped if this is the last call site.

caro3801 (changes requested):
The page-tracking work is a great improvement to the local search, the toolbox-offset rewrite to native `scrollIntoView` is a also good simplification. 
The main thing worth another look before merging is the finding about keyboard-driven "next document" path that can leave a stale pending page armed across a document switch, which is a real (if narrow) user-facing glitch. 
The keydown event firing while user is typing is a good one to fix alongside it since it shares the same `keydown` listener. 
Other comments are smaller polish, happy to see them handled now or later.

src/components/Document/DocumentViewer/DocumentViewerPdf.vue
pirhoo: Applied in 311623d7b. Confirmed the premise: `DocumentViewTabsViewer.vue:69` renders the viewer through `<component :is>` with no `:key`, so PDF-to-PDF navigation reuses the instance.

Not reproduced end to end though: this environment only has one PDF, so it rests on reading the code rather than a repro.

src/components/Document/DocumentViewer/DocumentViewerPdf.vue
pirhoo: Applied in 311623d7b, as a named `scrollsDocument()` predicate.

It was worse than described: holding an arrow key in the page field repeats `keydown`, so every repeat released the hold that the previous increment had just armed.

Injecting keys 250ms into a jump to page 150, distinct indicator values seen:

| injected | before | after |
|---|---|---|
| `" "` on the search input | `150, 25, 71, …, 149` | `150` |
| `ArrowDown` on the page field | `150, 9, 51, …, 149` | `150` |
| `PageDown` on `body` | `150, 37, 90, …, 149` | `150, 37, 90, …, 149` |

src/components/Document/DocumentViewer/DocumentViewerPdf.vue
pirhoo: Left as is, because this contradicts your earlier comment on the same tracker: that one asked the indicator to stop claiming the highlight page while a different page sits under the toolbar, and this one asks to set exactly that value on click.

They disagree often enough to matter. Stepping occurrences of "hedging", 2 of 5 steps had the active highlight on page 22 while page 21 was under the toolbar, so the indicator would show 22 on click and then visibly correct itself to 21 at settle.

It can be one or the other. I kept the "page under the toolbar" invariant. One line to flip if you would rather have the instant feedback.

src/components/Document/DocumentViewer/DocumentViewerPdf/DocumentViewerPdfPage.vue
pirhoo: Done now rather than as a follow-up, in 311623d7b: `block: 'center'` replaces the manual `-windowHeight / 2` offset, so `VueScrollTo`, `useScrollParent` and `useWindowSize` are all gone from this component. The highlight lands at `y=324` in a 630px viewport, against 316-331 before.

The dependency itself stays: `Core.js:187` registers it as a plugin and `DocumentEntriesList.vue:54` still calls it.

caro3801 (approved):
lgtm ! 🥳

PR #498, opened 2026-08-20
refactor(search): factor the fresh-stamp route push into a composable

lschim:
Follow-up to #496 (review suggestion): `SearchBar.vue`'s `submit()` and `useSearchFilter`'s `refreshRouteFromStart()` (used by the advanced search, filters, indices and search-operator changes) both push a fresh route to page one so the search always resubmits, but each rolled its own version — one regenerating a random stamp inline, the other going through the store's `refreshStamp()`/`toRouteQueryWithStamp`. Nothing kept them in sync, so the two could silently drift apart again.

Extracted the shared behavior into `useRefreshRouteFromStart()`, used by both. It takes the search store as a parameter instead of resolving it itself: some components use a suffixed/disposable store instance (`useSearchStore.inject()`) rather than the default one, so the composable must reuse the caller's own store reference rather than risk resolving a different one.

Refactoring made assisted by Claude

### DIFF
--- a/src/composables/useRefreshRouteFromStart.js
+++ b/src/composables/useRefreshRouteFromStart.js
@@ -1,0 +1,20 @@
+import { useRouter } from 'vue-router'
+
+/**
+ * Push a freshly stamped route to the first page of the search, so the search
+ * always resubmits even when the query and filters haven't changed.
+ *
+ * @param {object} searchStore - The search store instance to push a route for.
+ * @returns {{ refreshRouteFromStart: Function }}
+ */
+export function useRefreshRouteFromStart(searchStore) {
+  const router = useRouter()
+
+  function refreshRouteFromStart() {
+    searchStore.refreshStamp()
+    const query = { ...searchStore.toRouteQueryWithStamp, from: 0 }
+    return router.push({ name: 'search', query })
+  }
+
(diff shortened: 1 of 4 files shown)


src/composables/useRefreshRouteFromStart.js
```diff
@@ -0,0 +1,30 @@
...
+ * its own stamp or "from" reset logic.
+ *
+ * Takes the search store instance rather than resolving it itself: some
+ * components (batch search forms, linked documents) run against a suffixed,
+ * disposable store via `useSearchStore.inject()` instead of the default one,
+ * so the caller's own store reference must be reused as-is.
+ *
+ * @param {object} searchStore - The search store instance to push a route for.
+ * @returns {{ refreshRouteFromStart: Function }}
+ */
```
pirhoo, line 9: suggestion: could we trim this to the one non-obvious *why* (the stamp exists so an unchanged query still resubmits)? The rest restates the body and lists call sites that will drift as they change.

```suggestion
/**
 * Push a freshly stamped route to page one, so the search resubmits even when
 * the query and filters are unchanged.
 */
```

pirhoo (approved):

caro3801 (approved):
Thanks for this follow up. It's well tested and more clear

PR #500, opened 2026-08-21
feat(search): lock/unlock from the Filters panel

caro3801:
Part of the locked filters epic (icij/datashare#2219). Stacked PR 2/6, based on #499 (icij/datashare#2327).

Implements icij/datashare#2328: a lock toggle on Filters-panel value rows, wired to the store from #2327.

## What this PR does

Adds the first user-facing interaction of the locked-filters feature: a lock toggle on each value row
in the Filters panel, so a user can pin a specific filter value (e.g. `language = ENGLISH`) so it
survives "Clear filters" and gets carried across searches/projects, independent of whether it's
currently ticked.

**Key behavior:**
- **Lock/unlock a value from its row** - clicking the lock icon on a filter value locks it into the
  new personal, cross-project `lockedFiltersStore` (from #2327); locking an unticked value also ticks
  it (one click both applies and locks).
- **Locked-but-missing values stay visible** - if a locked value's real aggregation bucket disappears
  (a deleted tag, a re-indexed path), a synthetic zero-count "ghost" row keeps it visible and
  unlockable instead of silently vanishing from the panel. This synthesis is skipped while the panel's
  search box has a query, since real buckets are already filtered server-side and a synthetic row
  would otherwise outlive a search it never matched.
- **Unlock-on-removal, everywhere a value can be removed** - unticking a value, clearing all values
  for a filter, or removing/unregistering the filter itself all clean up that value's lock, so nothing
  is left stranded in the persisted store.
- **`lockable` opt-in gate** - the lock button only renders when a consumer explicitly passes
  `lockable` to `FiltersPanelSectionFilterEntry`, currently only `FilterType.vue`'s own entry. This
  keeps the button from leaking onto rows that never wire up `update:locked` (the "All" pseudo-row,
  disposable/unrelated contexts like the batch-search creation form).

Complex/paired filters are not part of this PR (path, category, starred, recommendedBy) - those get
lock support in #501 (`FilterTypeStarred`) and #504 (`FilterTypePath`, `FilterTypeRecommendedBy`). This
PR and #501 are tightly coupled - review them together if possible.

## Important files to review

- **`src/store/modules/lockedFilters.js`** - the store itself: entry shape (`{ name, value, label }`,
  mode baked into `name` via the `-` prefix), `lock`/`unlock`/`isLocked`, and the `unlockWhere`
  helper used by every removal path below.
- **`src/components/Filter/FilterType/FilterType.vue`** - wires the lock button to the store:
  `lockedName` (line 158, the store key for this filter's locks, now delegated to
  `useSearchFilter`'s shared `lockedNameFor` so it can't drift from the removal paths' own key),
  `isItemLocked`/`toggleLock` (180-190), and `missingLockedBucketsPage` (242) which synthesizes the
  ghost row for locked-but-missing values - the trickiest piece of logic in this PR, worth the closest
  look.
- **`src/components/FiltersPanel/FiltersPanelSectionFilterEntry.vue`** - the actual lock button/icon,
  `lockable`/`locked` props.
- **`src/composables/useSearchFilter.js`** - `removeFilterValue`/`removeFilterValues`/`removeFilter`,
  where unlock-on-removal is wired in for every value-removal path, and `lockedNameFor`, the single
  shared derivation of a filter's lock store key.
- **`src/store/modules/search.js`** - `removeFilter`'s own unlock fix (covers `FiltersMixin`'s
  `unregisterFilter` too, since both route through here), and the O(1)/`unlockWhere` perf follow-ups.

## Fixed from review

Addressed directly on this PR (see inline replies for detail):
1. The lock button showed up on every ticked row regardless of whether the parent wired
   `update:locked`, leaking a dead control onto "All" rows (ticked by default) and any future consumer
   that never hooks up locking. Fixed by gating the button behind an explicit `lockable` prop, passed
   only from `FilterType.vue`'s own entry.
2. `missingLockedBucketsPage` synthesized a locked-but-missing "ghost" row unconditionally, even while
   the panel's search box had an active query, so a locked value could keep appearing under a search it
   never matched (real buckets are filtered server-side and don't have this problem). Fixed by skipping
   synthesis entirely while a query is active, and split the computed into named steps along the way.
3. The count badge was hidden for every locked row, including a real value with a genuine document
   count, to work around the synthesized ghost rows (the only ones with a meaningless count). Fixed by
   giving synthetic buckets `doc_count: NaN` instead of `0`, so the existing `isNaN` guard masks only
   the ghost rows.
4. `lockedName` in `FilterType.vue` independently re-derived the same lock-store key that
   `useSearchFilter.js`'s `lockedNameFor` already computes for every removal path. Extracted
   `lockedNameFor` as the single shared source instead, so the lock button and the removal paths can't
   key the same row's lock differently.

## Known issues

Known issues, addressed elsewhere in the stack:
1. Flipping a filter's include/exclude mode orphans its existing locks (stale entry persists, unreachable from the UI). Fixed by commit 19960b817 on the stacked branch, landing with #2332 (#503) - that fix only covers a filter's currently *active* values; a locked-but-unticked value is instead caught by the `hasConflictingLocks`/`applyLockedFilters` mechanism, also #2332. Note this commit is not part of *this* PR's branch/ancestry - it addresses the epic as a whole, not #500 standalone. Relatedly, `removeFilterValue`/`removeFilterValues` only unlock the entry filed under the filter's *current* mode, so a value locked while excluded and later unticked/cleared after the filter flips back to include can leave the same kind of orphaned `-name` entry behind outside of a bare mode-flip - worth confirming #2332's fix covers this path too, not just the mode-flip itself.
2. `parseLockedName` was flagged as unused when this PR was first opened - no longer accurate, it's used directly by this PR's own `removeFilter` fix (#1 above).
3. ~~`FiltersPanelSectionFilterEntry.vue:69` - the lock button showed up whenever a row was ticked, with no opt-in gate (leaking a non-functional lock button onto "All" rows).~~ Fixed directly on this PR via the `lockable` opt-in prop (see "Fixed from review" #1 above) rather than deferred to #501.
4. `TaskBatchSearchFormFilters.vue` reuses `FilterType.vue`, which calls the global `useLockedFiltersStore()` internally, even though the batch-search form runs its own disposable, scoped search store - locking a value there leaks into the user's real, persisted lock store, since `FilterType.vue` currently passes `lockable` unconditionally on its own entry regardless of context. Still open, planned for #501/#504 - note that #501's current diff independently reintroduces the old `hideLock` opt-out prop to address this, predating this PR's move to `lockable`; it will need rebasing onto this branch and reconciling with `lockable` rather than stacking both mechanisms.
5. Baking include/exclude mode into a `-`-prefixed string on `name` instead of storing `excluded` as its own field meant a mode flip needed a delete-under-old-key/insert-under-new-key dance. Scoped down from a full storage-format rewrite (which would've touched every caller across the stack and changed the "distinct locks per mode" behavior an existing test relies on) to a smaller `lockedFiltersStore.retag({ name, newName, value })` - an in-place rename, used by #2332's `relockFilterValues`. Landed on #504.
6. `missingLockedBucketsPage` (`FilterType.vue`) synthesizes a zero-count "ghost" row for any locked value with no matching bucket in the current aggregation - meant to cover a genuinely deleted/re-indexed value, but it's built from only the currently-loaded infinite-scroll page, so a locked value that's real but not yet loaded renders as falsely "deleted" until the user scrolls further, and always renders above real buckets regardless of sort order. Still open, planned for #504: gate on `reachedBucketsEnd` (only decide a value is missing once every real bucket has loaded), and render the synthetic row after real buckets instead of before.

Nothing left open from this PR's own review at this point - every comment from the inline review round is fixed directly on this PR (see "Fixed from review" above); the remaining items are either already known and tracked in the epic's later PRs, or forward-looking notes for whoever picks those up next.

**Stack:** #2327 ← this PR ← #2329 ← #2330 ← #2331/#2332 ← #2336

### DIFF
--- a/src/components/Filter/FilterType/FilterType.vue
+++ b/src/components/Filter/FilterType/FilterType.vue
@@ -161,10 +161,10 @@ const hasAnyValue = computed(() => {
 })
 
 const toggleValue = async (item, checked) => {
+  // Unlocking on removal is handled centrally by useSearchFilter's
+  // removeFilterValue/removeFilterValues, so every removal path (this
+  // checkbox, the "All" toggle, breadcrumb chip removal) unlocks alike.
   await toggleFilterValue(filter, item, checked)
-  if (!checked) {
-    lockedFiltersStore.unlock({ name: lockedName.value, value: item.key })
-  }
   if (contextualize.value) {
     await aggregateOver()
   }
@@ -209,8 +209,13 @@ const excludedBucketsPage = computed(() => {
 // matching real aggregation bucket (a deleted tag, a re-indexed path).
 // Uses the exact same shape/insertion mechanism as excludedBucketsPage
 // above so it flows through bucketsWithExcludedValues unchanged.
+// Also excludes keys already synthesized by excludedBucketsPage: a value
+// that is ticked + excluded + contextualized is already rendered by that
+// mechanism regardless of locking, and de-duping here (rather than there)
+// keeps excludedBucketsPage's own semantics untouched.
 const missingLockedBucketsPage = computed(() => {
-  const realKeys = buckets.value.map(item => toString(item.key))
+  const excludedKeys = getPageBuckets(excludedBucketsPage.value).map(item => toString(item.key))
+  const realKeys = [...buckets.value.map(item => toString(item.key)), ...excludedKeys]
   const missing = lockedFiltersStore.entries
     .filter(entry => entry.name === lockedName.value && !realKeys.includes(entry.value))
     .map(entry => ({ key: entry.value, doc_count: 0, __lockedLabel: entry.label }))
--- a/src/components/Filter/FilterType/FilterType.vue
+++ b/src/components/Filter/FilterType/FilterType.vue
@@ -21,6 +21,7 @@ import FiltersPanelSectionFilterEntry from '@/components/FiltersPanel/FiltersPan
 import FilterTypeAll from '@/components/Filter/FilterType/FilterTypeAll'
 import settings from '@/utils/settings'
 import { useSearchStore, useLockedFiltersStore } from '@/store/modules'
+import { toLockedName } from '@/store/modules/lockedFilters'
 import builtinFilterIcons from '@/store/filters/icons'
 
 const query = defineModel('query', { type: String, default: '' })
@@ -148,7 +149,10 @@ const exclude = computedExcludeFilter(filter)
 // The store key for this filter's locks: the filter's own current
 // include/exclude mode, never a paired dimension's — locking a chip on one
 // side of a paired filter must never lock or affect the other side.
-const lockedName = computed(() => (exclude.value ? `-${filter.name}` : filter.name))
+// TODO(locked-filters): filter types overriding FilterType's default slot
+// (FilterTypeFileTypes, FilterTypePath, FilterTypeProject, FilterTypeStarred,
+// FilterTypeRecommendedBy) don't receive lock/unlock support yet — see follow-up issue.
+const lockedName = computed(() => toLockedName(filter.name, exclude.value))
 const sort = computedSortFilter(filter)
 const contextualize = computedContextualizeFilter(filter)
 
@@ -184,6 +188,9 @@ function toggleLock(item, locked) {
 }
 
 const bucketLabel = (bucket) => {
+  // The label is frozen at lock time and intentionally does not re-translate
+  // if the UI language changes later — the bucket is gone, so there's
+  // nothing left to re-resolve the label against.
   if (bucket?.__lockedLabel != null) {
     return bucket.__lockedLabel
   }
@@ -213,6 +220,10 @@ const excludedBucketsPage = computed(() => {
 // that is ticked + excluded + contextualized is already rendered by that
 // mechanism regardless of locking, and de-duping here (rather than there)
 // keeps excludedBucketsPage's own semantics untouched.
+// Like excludedBucketsPage above, this does not filter against the panel's
+// search box: a locked-but-missing value always stays visible regardless of
+// what the user is searching for. Consistent with existing behavior, not a
+// new gap.
 const missingLockedBucketsPage = computed(() => {
   const excludedKeys = getPageBuckets(excludedBucketsPage.value).map(item => toString(item.key))
   const realKeys = [...buckets.value.map(item => toString(item.key)), ...excludedKeys]
--- a/src/components/Filter/FilterType/FilterType.vue
+++ b/src/components/Filter/FilterType/FilterType.vue
@@ -7,6 +7,7 @@ import flatten from 'lodash/flatten'
 import get from 'lodash/get'
 import noop from 'lodash/noop'
 import setWith from 'lodash/setWith'
+import toString from 'lodash/toString'
 import uniqueId from 'lodash/uniqueId'
 import InfiniteLoading from 'v3-infinite-loading'
 import { useI18n } from 'vue-i18n'
@@ -19,7 +20,7 @@ import FiltersPanelSectionFilter from '@/components/FiltersPanel/FiltersPanelSec
 import FiltersPanelSectionFilterEntry from '@/components/FiltersPanel/FiltersPanelSectionFilterEntry'
 import FilterTypeAll from '@/components/Filter/FilterType/FilterTypeAll'
 import settings from '@/utils/settings'
-import { useSearchStore } from '@/store/modules'
+import { useSearchStore, useLockedFiltersStore } from '@/store/modules'
 import builtinFilterIcons from '@/store/filters/icons'
 
 const query = defineModel('query', { type: String, default: '' })
@@ -55,6 +56,7 @@ const expand = ref(false)
 
 const { waitFor, isLoading } = useWait({ throttle: 500 })
 const searchStore = useSearchStore.inject()
+const lockedFiltersStore = useLockedFiltersStore()
 
 const aggregateOver = () => {
   return aggregate({ clearPages: true })
@@ -143,6 +145,10 @@ const {
 } = useSearchFilter()
 
 const exclude = computedExcludeFilter(filter)
+// The store key for this filter's locks: the filter's own current
+// include/exclude mode, never a paired dimension's — locking a chip on one
+// side of a paired filter must never lock or affect the other side.
+const lockedName = computed(() => (exclude.value ? `-${filter.name}` : filter.name))
 const sort = computedSortFilter(filter)
 const contextualize = computedContextualizeFilter(filter)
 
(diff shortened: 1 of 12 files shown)


src/components/FiltersPanel/FiltersPanelSectionFilterEntry.vue
```diff
@@ -54,7 +63,12 @@ const classList = computed(() => {
   }
 })
 
-const showCount = computed(() => !props.hideCount && !isNaN(props.count))
+// Show the lock button when the value is ticked OR already locked: a value
+// can be locked while unticked (e.g. after "Clear filters", which preserves
+// locks but unticks the value), and the user still needs a way to unlock it.
+const showLockButton = computed(() => Boolean(props.modelValue) || props.locked)
```
pirhoo: important: gating on `modelValue || locked` alone opts every consumer of this shared entry into a
lock button, but only `FilterType.vue` wires `@update:locked`, so `FilterTypeAll`,
`FilterTypeProjectAll`, `ContentTypesAll` and `ContentTypesCategoryName` render a dead control on
any ticked row, and "All" rows are ticked by default. Known issue 3 points at `hideLock` in #501,
but that lands as opt-out and `FilterTypeProjectAll` / `ContentTypesAll` still pass nothing even at
the stack tip, so the leak survives the whole stack. Could we make it opt-in instead?

```suggestion
const showLockButton = computed(() => props.lockable && (Boolean(props.modelValue) || props.locked))
```

That needs a `lockable: { type: Boolean, default: false }` prop next to `locked`, passed only from
the `filters-panel-section-filter-entry` in `FilterType.vue`. Consumers that want locking then say
so, and new ones can't inherit the button by accident.

src/components/Filter/FilterType/FilterType.vue
```diff
@@ -180,8 +213,29 @@ const excludedBucketsPage = computed(() => {
...
+// what the user is searching for. Consistent with existing behavior, not a
+// new gap.
+const missingLockedBucketsPage = computed(() => {
+  const excludedKeys = getPageBuckets(excludedBucketsPage.value).map(item => toString(item.key))
+  const realKeys = [...buckets.value.map(item => toString(item.key)), ...excludedKeys]
+  const missing = lockedFiltersStore.entries
+    .filter(entry => entry.name === lockedName.value && !realKeys.includes(entry.value))
+    .map(entry => ({ key: entry.value, doc_count: 0, __lockedLabel: entry.label }))
+  return setWith({}, pageBucketsPath.value.join('.'), missing, Object)
+})
```
pirhoo: important: the panel's search box filters real buckets server-side through
`aggregationOptions.include`, but the synthesized rows never see `query`, so typing `engl` on
`language` with FRENCH locked still lists FRENCH, and prepending puts it above the matching
ENGLISH row. The parity note with `excludedBucketsPage` doesn't quite carry, since that path only
fires on the rare `contextualize && exclude` combo while this one fires for every locked value.
Splitting the one dense computed into named steps makes the fix a single guard clause:

```suggestion
// A locked value with no matching real bucket gets a synthetic zero-count row so it
// stays visible and unlockable (a deleted tag, a re-indexed path). Skipped while the
// panel's search box is active: real buckets are filtered server-side via
// aggregationOptions.include, so a synthetic row would outlive a query it never matched.
const bucketKey = bucket => toString(bucket.key)

const renderedBucketKeys = computed(() => {
  const realKeys = buckets.value.map(bucketKey)
  // excludedBucketsPage already synthesizes the ticked + excluded + contextualized case,
  // so de-dupe here and leave that computed's own semantics untouched.
  const excludedKeys = getPageBuckets(excludedBucketsPage.value).map(bucketKey)
  return new Set([...realKeys, ...excludedKeys])
})

const missingLocks = computed(() => {
  const isForThisFilter = entry => entry.name === lockedName.value
  const isMissing = entry => !renderedBucketKeys.value.has(entry.value)
  return lockedFiltersStore.entries.filter(entry => isForThisFilter(entry) && isMissing(entry))
})

const toSyntheticBucket = entry => ({ key: entry.value, doc_count: 0, __lockedLabel: entry.label })

const missingLockedBucketsPage = computed(() => {
  if (query.value !== '') {
    return []
  }
  const missing = missingLocks.value.map(toSyntheticBucket)
  return setWith({}, pageBucketsPath.value.join('.'), missing, Object)
})
```

Three things fall out of the split beyond the query guard. `renderedBucketKeys` as a `Set` drops
the `realKeys.includes` scan to O(1) per lock, in the same spirit as this PR's own `indexByKey`
commit. `missingLocks` becomes independently readable and testable, which is where the
locked-but-not-yet-loaded case from known issue 6 will want to live. And the 12-line comment
shrinks to the two *why*s that the names can't carry.

Returning `[]` matches what `excludedBucketsPage` already does in its negative case, so
`getPageBuckets` handles it unchanged. Filtering `missingLocks` against `queryTokens` instead
would work too, if you'd rather keep locked values visible while searching.

src/components/FiltersPanel/FiltersPanelSectionFilterEntry.vue
```diff
@@ -54,7 +63,12 @@ const classList = computed(() => {
...
+const showCount = computed(() => !props.hideCount && !isNaN(props.count) && !props.locked)
```
pirhoo: suggestion: `!props.locked` hides the badge for every locked row, so locking a real value with
1,234 documents drops a genuine count to work around the synthesized ghost rows, which are the
only ones carrying a meaningless zero. Giving the synthesized buckets `doc_count: NaN` instead of
`0` in `missingLockedBucketsPage` lets the existing `isNaN` guard mask exactly those rows and drops
the clause here:

```suggestion
const showCount = computed(() => !props.hideCount && !isNaN(props.count))
```

The four count assertions in the new spec would move with it, and `locked` stops being a display
concern for this component.

pirhoo (changes requested):
Thanks, the PR description and its known-issues list made a stacked branch much easier to review.

Three inline comments below. Label legend: `important:` a real defect, I'd expect a change but happy to discuss it; `suggestion:` cleanup, accept or not is up to you.

src/components/FiltersPanel/FiltersPanelSectionFilterEntry.vue
caro3801: Done in baebb8773. Locking is opt-in now via a `lockable` prop passed only from FilterType.vue's own
entry, so a consumer that never wires `update:locked` can no longer inherit a dead lock button on its
ticked rows.

src/components/Filter/FilterType/FilterType.vue
caro3801: Done in 0a990bcfc. Real buckets get filtered server-side through `aggregationOptions.include` once the
panel's search box has a query, so the synthesized locked-but-missing rows now skip themselves while a
query is active instead of outliving a search they never matched. Split into the named steps you
suggested.

src/components/FiltersPanel/FiltersPanelSectionFilterEntry.vue
caro3801: Done in ed397cd61. The synthetic ghost buckets now carry `doc_count: NaN` instead of 0, so the existing
`isNaN` guard hides exactly those and a real locked value keeps showing its genuine count.

pirhoo (approved):

PR #501, opened 2026-08-21
feat(search): breadcrumb lock + auto-apply locks 

caro3801:
Part of the locked filters epic (icij/datashare#2219). Stacked PR 3/6, based on #500 (#2328).

Implements icij/datashare#2329: the core mechanic of the epic — a lock toggle on breadcrumb chips, and silent auto-merge of locked values into the live search on every route hydration. Also extends lock support beyond the Filters-panel's default slot to the `starred` and `contentType` filters.

## What this PR does

- **Lock/unlock from a breadcrumb chip** - each removable breadcrumb chip (except the free-text query chip and the project chip) gets its own lock toggle, wired to the same personal `lockedFiltersStore` the Filters panel uses. Suppressed on read-only chips (`noXIcon`), since those render disposable/persisted queries (saved searches, batch overviews) that must never mutate the user's real lock store.
- **Silent auto-merge on route hydration** - `updateFromRouteQuery` resets `values`/`excludeFilters` on every route entry (including the first search of a session, e.g. opening a shared link), so locks are re-merged into the live state every time, not just once at lock-creation. A lock only yields to an explicit route conflict (e.g. route says `f[-contentType]`, lock is for included `contentType`) - resolving that conflict is icij/datashare#2332's job. Disposable/read-only hydrations (`SearchBreadcrumbUri.vue`) opt out via a new `mergeLocks: false` option, so a saved search's chips never pick up the viewer's personal locks.
- **Lock support for `starred` and `contentType`** - both override `FilterType.vue`'s default slot and needed their own `isItemLocked`/`toggleLock` wiring (`contentType`'s category-grouped view keeps this from PR #2329, individual leaf types only here).

## Important files to review

- **`src/store/modules/search.js`** - `updateFromRouteQuery`'s new `mergeLocks` option and `mergeLockedFilters()` (conflict detection, paired-dimension handling). The trickiest logic in this PR.
- **`src/components/Search/SearchBreadcrumbForm/SearchBreadcrumbFormEntry.vue`** - the breadcrumb lock toggle: `isLockable`, `locked`/`lockedForDisplay`, and the humanized-label resolution in `toggleLock`.
- **`src/components/Search/SearchParameter/SearchParameterQueryTerm.vue`** - the chip's lock icon markup/keyboard accessibility.
- **`src/components/Filter/FilterType/FilterTypeStarred.vue`** and **`src/components/Filter/FilterType/FilterTypeFileTypes.vue`** - lock wiring for the two filter types that override `FilterType.vue`'s default slot, plus `hideLock` threading in the latter.
- **`src/components/Search/SearchBreadcrumbUri/SearchBreadcrumbUri.vue`** - the `mergeLocks: false` opt-out for read-only/disposable breadcrumb displays.

## Known issues, addressed elsewhere in the stack

- Flipping a filter's include/exclude mode orphans its existing locks (stale entry persists, unreachable from the UI) - fixed on #2332.
- `FilterTypePath`, `FilterTypeProject`, `FilterTypeRecommendedBy` still don't have lock support - #2336.

**Stack:** #2327 ← #2328 ← this PR ← #2330 ← #2331/#2332 ← #2336


src/components/Filter/FilterType/FilterType.vue
```diff
@@ -38,6 +38,15 @@ const { filter, modal, hideCount, overlayShow } = defineProps({
...
   },
+  // Suppresses the per-value lock button and locked-but-missing synthetic
+  // buckets entirely. Used by disposable/unrelated screens (e.g. the batch
+  // search creation form) that render this filter against a non-live search
+  // store: locking from there would write to the user's global personal lock
+  // store with no visibility into what it's actually affecting. See
+  // icij/datashare#2329.
+  hideLock: {
+    type: Boolean
+  },
```
pirhoo, line 48: crucial: `hideLock` suppresses the button but not the writes, so the batch-search form still mutates the user's real lock store. Unticking a value there goes `toggleValue` -> `useSearchFilter.removeFilterValue` -> `lockedFiltersStore.unlock(...)`, and clicking "All" goes `removeFilterValues` -> `unlockWhere(entry.name === lockedName)`, wiping every lock for that filter. Could we thread the flag down, either by skipping the unlock in the removal paths when `hideLock` is set, or by giving `removeFilterValue`/`removeFilterValues` an opt-out that this component passes through?

src/components/Filter/FilterType/FilterTypeFileTypes.vue
```diff
@@ -52,11 +61,37 @@ const {
...
+    // applies and locks the filter.
+    if (!hasFilterValue(props.filter, { key: contentType })) {
+      await toggleFilterValue(props.filter, { key: contentType }, true)
+    }
+    lockedFiltersStore.lock({ name: lockedName.value, value: contentType, label: getDocumentTypeLabel(contentType) })
+  }
+  else {
+    lockedFiltersStore.unlock({ name: lockedName.value, value: contentType })
+  }
+}
```
pirhoo: important: the auto-select goes through the raw `toggleFilterValue` instead of `toggleEntry`, so it skips the contentType/contentTypeCategory promote/demote lockstep. With `f[contentTypeCategory]=DOCUMENT` stored, `hasFilterValue` is false for `application/pdf`, so locking it appends `f[contentType]=application/pdf` on top of the stored category, which is the state `uncheckChildOfStoredCategory` documents as a URL-tamper / race condition. `isEntryRetainedDuringSearch` is the category-aware "already applied" check, and `toggleEntry` is synchronous:

```suggestion
function toggleLock(contentType, locked) {
  if (locked) {
    // Locking an unticked value also selects it — a single click both
    // applies and locks the filter.
    if (!isEntryRetainedDuringSearch(contentType)) {
      toggleEntry(contentType, true)
    }
    lockedFiltersStore.lock({ name: lockedName.value, value: contentType, label: getDocumentTypeLabel(contentType) })
  }
  else {
    lockedFiltersStore.unlock({ name: lockedName.value, value: contentType })
  }
}
```

src/composables/useContentTypeSelection.js
```diff
@@ -187,6 +198,10 @@ export function useContentTypeSelection({ filter, categories }) {
   const dropCategoryAnd = (category, keepFromCategory) => {
     const categoryTypes = typesInCategory(category)
     const others = currentContentTypes().filter(value => !categoryTypes.includes(value))
+    // Unlock every type this category covered except the ones the caller
+    // asks to keep explicit (the clicked child on demote, the surviving
+    // siblings on uncheck-with-stored-category).
+    categoryTypes.filter(type => !keepFromCategory.includes(type)).forEach(unlockContentType)
```
pirhoo, line 204: important: `unlockContentType` landed here and in `uncheckPlainChild`, but `promoteToCategory` just below also strips explicit children and was missed. A locked child keeps its lock after promotion, so the next hydration re-adds `f[contentType]=application/pdf` beside `f[contentTypeCategory]=DOCUMENT`. Same one-liner there:

```js
const promoteToCategory = (category) => {
  const categoryTypes = typesInCategory(category)
  categoryTypes.forEach(unlockContentType)
  writeContentTypes(currentContentTypes().filter(value => !categoryTypes.includes(value)))
  writeCategories([...currentCategories(), category])
}
```

src/composables/useContentTypeSelection.js
```diff
@@ -238,6 +253,7 @@ export function useContentTypeSelection({ filter, categories }) {
       writeCategories([...remainingCategories, category])
       return
     }
+    types.forEach(unlockContentType)
```
pirhoo: important: the checked branch returns before reaching this line, yet it also strips the explicit children, so promoting a category leaves their locks behind and the next hydration re-adds them alongside the stored category. Hoisting this call above the `writeContentTypes` call at the top of `toggleCategory` covers both branches.

src/components/FiltersPanel/FiltersPanelSectionFilterEntry.vue
```diff
@@ -63,11 +66,11 @@ const classList = computed(() => {
...
 
-// Show the lock button when the value is ticked OR already locked: a value
-// can be locked while unticked (e.g. after "Clear filters", which preserves
-// locks but unticks the value), and the user still needs a way to unlock it.
-const showLockButton = computed(() => Boolean(props.modelValue) || props.locked)
-const showCount = computed(() => !props.hideCount && !isNaN(props.count) && !props.locked)
+// The lock button's slot is always reserved (visibility handled purely via
+// CSS opacity — hidden by default, revealed on hover or when locked) so the
+// count pill's position never jitters based on tick/lock state.
+const showLockButton = computed(() => !props.hideLock)
```
pirhoo: important: dropping the `modelValue || locked` condition means the padlock now renders on every entry using this component, including the ones that never bind `@update:locked`: `ContentTypesAll.vue`, `ContentTypesCategoryName.vue`, `FilterTypeProject.vue`, `FilterTypeProjectAll.vue` and `FilterTypeRecommendedBy.vue` (verified by mounting the first two, the button is there and inert). `FilterTypeAll.vue` got `hide-lock` in this diff, could those five get it too until #2336 wires them up?

src/components/FiltersPanel/FiltersPanelSectionFilterEntry.vue
```diff
@@ -147,13 +154,38 @@ const lockLabel = computed(() => t(props.locked ? 'filtersPanelSectionFilterEntr
...
+      &:deep(.button-icon__icon-left) {
+        color: var(--bs-action);
+      }
+    }
+  }
+
+  &:hover &__lock,
+  &__lock:focus-visible {
+    opacity: 1;
   }
```
pirhoo, line 196: important: `opacity: 0` still hit-tests, and touch devices never match `:hover`, so on mobile every filter row carries an invisible tap target that silently creates a lock. Gating the hover-reveal on `hover: hover` keeps the desktop behavior and the keyboard focus ring, while leaving the button visible where hover does not exist:

```suggestion
  &__lock {
    flex-shrink: 0;
    transition: opacity 0.15s ease;

    // Hover-reveal only where hovering exists: on touch the row never
    // matches :hover, so an opacity-0 lock stays an invisible tap target.
    @media (hover: hover) {
      opacity: 0;
    }

    &--locked {
      opacity: 1;

      // Same accent as the breadcrumb chip's locked lock icon.
      &:deep(.button-icon__icon-left) {
        color: var(--bs-action);
      }
    }
  }

  @media (hover: hover) {
    &:hover &__lock,
    &__lock:focus-visible {
      opacity: 1;
    }
  }
```

src/components/Search/SearchParameter/SearchParameterQueryTerm.vue
```diff
@@ -102,6 +117,19 @@ const showOperator = computed(() => {
...
+      role="button"
+      tabindex="0"
+      class="search-parameter-query-term__lock"
+      :aria-pressed="locked"
+      :aria-label="lockLabel"
+      :name="locked ? IPhLock : IPhLockOpen"
+      @click="emit('click:lock')"
+      @keydown.enter.prevent="emit('click:lock')"
+      @keydown.space.prevent="emit('click:lock')"
+    />
```
pirhoo, line 133: important: `tag="span"` makes BButton emit `role="button"` on the chip root (its `nonStandardTag` branch), and ARIA treats the children of a `role="button"` as presentational, so this lock's `aria-label` and `aria-pressed` are never exposed to assistive tech. Meanwhile the outer `role="button"` gets no tabindex and has no action of its own. Could we render the lock as a sibling of the chip rather than inside it, or drop the implicit role on the root? Worth noting the new spec only asserts the root is not a native `<button>`, so it stays green either way.

pirhoo (changes requested):
Great, thanks! This is a big move. I've made several feedbacks, all blocking.

Reviewed only the stack-local diff (`feat/2219/2328-panel-lock...feat/2219/2329-breadcrumb-lock`). Labels: `crucial:` blocks merge, `important:` real defect, expect a change.

pirhoo (approved):

src/components/Filter/FilterType/FilterType.vue
caro3801: Done in 060d95cf. hideLock was meant to keep the batch-search form's disposable rendering from touching the user's real personal lock store, but only the button visibility was gated on it; unticking a value or clicking "All" still ran through useSearchFilter's unconditional unlock/unlockWhere. Added a skipUnlock option threaded from hideLock through toggleFilterValue/removeFilterValue/removeFilterValues/computedAll and down into FilterTypeAll.

src/components/Filter/FilterType/FilterTypeFileTypes.vue
caro3801: Done in c182b760, using your suggested diff. Locking now goes through toggleEntry/isEntryRetainedDuringSearch, so auto-selecting an unticked value stays in lockstep with the contentType/contentTypeCategory promote/demote instead of bypassing it.

src/composables/useContentTypeSelection.js
caro3801: Done in c182b760, using your suggested one-liner. promoteToCategory now unlocks the category's explicit children before stripping them, same as dropCategoryAnd's demote/uncheck paths.

src/composables/useContentTypeSelection.js
caro3801: Done in c182b760. Hoisted the unlock call to the top of toggleCategory so it runs before writeContentTypes strips the children, covering the checked branch too.

src/components/FiltersPanel/FiltersPanelSectionFilterEntry.vue
caro3801: Done in 8d6a755d (and 3d818762 after a rebase renamed the underlying gate to an opt-in `lockable` prop). ContentTypesAll, ContentTypesCategoryName, FilterTypeProject, FilterTypeProjectAll and FilterTypeRecommendedBy never opt into it, so none of them renders a lock button with nothing wired to it.

src/components/FiltersPanel/FiltersPanelSectionFilterEntry.vue
caro3801: Done in 283c9f7b, using your suggested diff. The hover-reveal is now gated on @media (hover: hover), so touch devices show the lock button plainly instead of carrying an invisible always-there tap target.

src/components/Search/SearchParameter/SearchParameterQueryTerm.vue
caro3801: Done in a3c5fef4, using the drop-the-implicit-role option. The lock icon stays nested inside the chip (matching the existing visual design, lock next to the remove "x"), but the chip's own role is now pinned to "presentation" instead of BButton's auto-assigned "button", so the lock's own role/aria-pressed/aria-label are exposed to assistive tech instead of being flattened into the chip's accessible name.

PR #502, opened 2026-08-21
feat(search): Clear filters preserves locks + Unlock filters

caro3801:
Part of the locked filters epic (icij/datashare#2219). Stacked PR 4/6, based on #501 (#2329).

Implements icij/datashare#2330: "Clear filters" now preserves locked values while removing unlocked ones, and adds an "Unlock filters (N)" breadcrumb footer action to drop every lock without touching applied filter values.

**Stack:** #2329 ← this PR ← #2331/#2332 ← #2336

**Known issue (fixed in #503):** `toggleFilter` in the search store only flips `excludeFilters` here; it doesn't yet re-tag locked entries when a filter's include/exclude mode changes. Since a lock's storage key is mode-encoded (`contentType` vs `-contentType`), locking a value then switching that filter to exclude mode (or back) makes the existing lock entry mismatch the new key, so it silently reads as unlocked in the breadcrumb/checkbox even though the entry itself is still stored. This is addressed in #503 via `relockFilterValues`, which re-tags the entry under the new key on mode flip. Not fixed here, since this PR predates that change in the stack.

### DIFF
--- a/src/composables/useSearchBreadcrumb.js
+++ b/src/composables/useSearchBreadcrumb.js
@@ -186,11 +186,19 @@ export function useSearchBreadcrumb() {
     return refreshRoute()
   }
 
-  const clearFiltersEntries = () => {
+  const clearFilterState = () => {
     searchStore.resetFilterValues()
+    // Exclusion mode is part of "filters" too: leaving it behind would re-apply
+    // an *included* lock in exclude mode, since resetFilterValues() wipes only
+    // `values` and mergeLockedFilters()'s conflict check keys off `values`.
+    searchStore.excludeFilters.slice().forEach(name => searchStore.includeFilter(name))
     // Locked values must survive a filter-wipe — re-merge them immediately
     // rather than waiting for the next route hydration. See icij/datashare#2330.
     searchStore.mergeLockedFilters()
+  }
+
+  const clearFiltersEntries = () => {
+    clearFilterState()
     return refreshRoute()
   }
 
@@ -200,8 +208,7 @@ export function useSearchBreadcrumb() {
   }
 
   const clearAll = () => {
-    searchStore.resetFilterValues()
-    searchStore.mergeLockedFilters()
+    clearFilterState()
     searchStore.resetQuery()
     return refreshRoute()
   }
--- a/src/composables/useSearchBreadcrumb.js
+++ b/src/composables/useSearchBreadcrumb.js
@@ -9,7 +9,7 @@ import unset from 'lodash/unset'
 import lucene from 'lucene'
 
 import { useSearchFilter } from '@/composables/useSearchFilter'
-import { useSearchBreadcrumbStore, useSearchStore } from '@/store/modules'
+import { useLockedFiltersStore, useSearchBreadcrumbStore, useSearchStore } from '@/store/modules'
 import { getCanonicalDimension, getPairedDimension } from '@/store/filters/pairedDimensions'
 import findPath from '@/utils/findPath'
 
@@ -109,6 +109,8 @@ export function useSearchBreadcrumb() {
   const searchStore = useSearchStore.inject()
   const searchBreadcrumbStore = useSearchBreadcrumbStore()
   const { setQuery, removeIndex, removeFilterValue, refreshRoute } = useSearchFilter()
+  const lockedFiltersStore = useLockedFiltersStore()
+  const lockedFiltersCount = computed(() => lockedFiltersStore.count)
 
   const count = computed(() => entries.value.length)
 
@@ -186,6 +188,9 @@ export function useSearchBreadcrumb() {
 
   const clearFiltersEntries = () => {
     searchStore.resetFilterValues()
+    // Locked values must survive a filter-wipe — re-merge them immediately
+    // rather than waiting for the next route hydration. See icij/datashare#2330.
+    searchStore.mergeLockedFilters()
     return refreshRoute()
   }
 
@@ -196,8 +201,13 @@ export function useSearchBreadcrumb() {
 
   const clearAll = () => {
     searchStore.resetFilterValues()
+    searchStore.mergeLockedFilters()
     searchStore.resetQuery()
     return refreshRoute()
+  }
+
+  const unlockAll = () => {
+    lockedFiltersStore.unlockAll()
   }
 
   return {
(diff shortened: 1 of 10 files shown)


src/composables/useSearchBreadcrumb.js
```diff
@@ -195,11 +211,15 @@ export function useSearchBreadcrumb() {
...
   const clearAll = () => {
-    searchStore.resetFilterValues()
+    clearFilterState()
     searchStore.resetQuery()
     return refreshRoute()
   }
 
+  const unlockAll = () => {
+    lockedFiltersStore.unlockAll()
+  }
```
pirhoo, line 213: important: looks like `unlockAll` never pushes the route, so values that only lived in the store because `mergeLockedFilters` merged them (never in the URL) stay visible and applied, then vanish on the next hydration (reload, or document then Back). Could we refresh the route the way the other actions here do?

```suggestion
  const unlockAll = () => {
    lockedFiltersStore.unlockAll()
    return refreshRoute()
  }
```

The per-chip padlock has the same shape, so it may be worth fixing at that call site too rather than only here.

src/composables/useSearchBreadcrumb.js
```diff
@@ -211,6 +231,8 @@ export function useSearchBreadcrumb() {
     clearFiltersEntries,
     clearQueryEntries,
     clearAll,
+    unlockAll,
+    lockedFiltersCount,
     count,
     hasQueryEntries,
     hasFiltersEntries,
```
pirhoo, line 230: important: `hasFiltersEntries` still counts locked chips, so "Clear filters" and "Clear filters and query" stay enabled after a lock-preserving clear, and further clicks are silent no-ops (same `toRouteQuery`, so duplicated navigation, no hydration, no refresh, chips unchanged). Could we count only the entries that would actually go away? The entry's `filter` already carries the `-` prefix, which matches the lock name format:

```js
  const unlockedFiltersEntries = computed(() => {
    return filtersEntries.value.filter(({ filter, value }) => !lockedFiltersStore.isLocked({ name: filter, value }))
  })
  const hasFiltersEntries = computed(() => !!unlockedFiltersEntries.value.length)
```

(The definition sits just above the diff, next to `hasQueryEntries`, so I could not attach a suggestion block to it.)

tests/unit/specs/composables/useSearchBreadcrumb.spec.js
```diff
@@ -99,4 +99,109 @@ describe('useSearchBreadcrumb composable', () => {
...
+      // left included even though the group must stay in lockstep.
+      expect(searchStore.isFilterExcluded('contentType')).toBe(true)
+      expect(searchStore.isFilterExcluded('contentTypeCategory')).toBe(true)
+    })
+
+    it('clearFiltersEntries removes everything when there are zero locks', () => {
+      // Locks persist to localStorage (persist: true) across the jsdom-shared
+      // localStorage instance, so a fresh pinia alone doesn't guarantee zero
+      // locks here — clear explicitly.
+      lockedFiltersStore.unlockAll()
```
pirhoo: important: the comment blames `persist: true` and localStorage, but the actual leak is that `CoreSetup.useAll()` installs the module-level pinia singleton, so every test in this file shares one lockedFilters store. That means the paired-dimension test above runs with a leftover include lock from an earlier test and passes for the wrong reason. Could we drop the explanation here and reset in the outer `beforeEach` instead, so the whole file starts clean?

```suggestion
      lockedFiltersStore.unlockAll()
```

src/composables/useSearchBreadcrumb.js
```diff
@@ -184,8 +186,22 @@ export function useSearchBreadcrumb() {
...
+    // `values` and mergeLockedFilters()'s conflict check keys off `values`.
+    searchStore.excludeFilters.slice().forEach(name => searchStore.includeFilter(name))
+    // Locked values must survive a filter-wipe — re-merge them immediately
+    // rather than waiting for the next route hydration. See icij/datashare#2330.
+    searchStore.mergeLockedFilters()
+    // mergeLockedFilters only merges values; a paired dimension's exclude mode
+    // still needs the same reconciliation updateFromRouteQuery always runs.
+    searchStore.reconcilePairedExcludeFilters()
+  }
+
```
pirhoo: suggestion: `clearFilterState` re-implements `updateFromRouteQuery`'s four-step invariant in the composable, which is why `mergeLockedFilters` and `reconcilePairedExcludeFilters` had to become public. That step list has already grown twice (icij/datashare#2329, icij/datashare#2351) and will drift again. A single `resetFilterValuesPreservingLocks()` action next to `resetFilterValues` in the search store would keep both internals private and give the composable one call.

pirhoo (changes requested):
Thanks, very neat!

src/composables/useSearchBreadcrumb.js
caro3801: Done in c1be28968. A value that only lived in the store because a lock silently merged it in, never in the URL, kept showing as applied after unlocking but would vanish on the next hydration (reload, or document then Back). Refreshing the route, like every other action in this composable already does, pushes it into the URL so it survives.

The per-chip padlock does not need the same fix: a displayed chip is always already route-backed (it comes from the breadcrumb's own query history), so toggling its lock does not add or remove any filter value the way unlockAll's bulk unlock does. There is no state to persist there.

src/composables/useSearchBreadcrumb.js
caro3801: Done in e0780061f. hasFiltersEntries counted locked chips too, so "Clear filters" and "Clear filters and query" stayed enabled after a lock-preserving clear even though nothing left would change, making a further click a silent no-op. It now counts only the entries that would actually go away.

tests/unit/specs/composables/useSearchBreadcrumb.spec.js
caro3801: Done in 32014481d. CoreSetup.useAll() reuses the module-level pinia singleton rather than a fresh instance per test, so lockedFiltersStore state was leaking across every test in this file; the paired-dimension test above was passing with a leftover lock for the wrong reason. Resetting once in the outer beforeEach starts the whole file clean instead of blaming persist/localStorage in one inner test's comment.

src/composables/useSearchBreadcrumb.js
caro3801: Done in eda5266c4. resetFilterValuesPreservingLocks() now lives next to resetFilterValues() in the search store, so mergeLockedFilters and reconcilePairedExcludeFilters are private again and the composable makes a single call instead of re-running the reset/merge/reconcile sequence itself.

pirhoo (approved):
Cool, thanks for the changes!

caro3801:
Thanks for the review(s)!

PR #503, opened 2026-08-21
feat(search): shared-link conflict handling + Apply locked filters icij/datashare#2331 icij/datashare#2332

caro3801:
Part of the locked filters epic (icij/datashare#2219). Stacked PR 5/6, based on #502 (#2330).

Covers two related, chronologically interleaved stories:

- icij/datashare#2331 (exclude locked filters from outward artifacts): implemented, then reversed in the same branch after deciding stripping locks from a saved search/batch task made the saved query silently narrower than what the user saw on screen — locks are meant to auto-apply everywhere, full stop. What survives from #2331 instead: opening a **saved search** now skips the hydration-time lock merge for that one navigation, since a saved search is a frozen snapshot.
- icij/datashare#2332 (Apply locked filters action): the breadcrumb footer's "Apply locked filters" button for the shared-link conflict case (an incoming query that disagrees with a lock's include/exclude mode), force-refresh, and auto-opening the breadcrumb panel after an explicit search submission when locks are active.

Also includes a small unrelated-but-discovered test-isolation fix (`lockedFiltersStore` reset moved to a global `beforeEach` instead of per-spec-file, since Pinia is a true module-level singleton in this app's test setup).

**Stack:** #2330 ← this PR ← #2336


src/views/Search/Search.vue
pirhoo, line 1: important: `handleAdvancedSearch` never calls `markJustSubmitted()`, unlike `submit()` in SearchBar.vue, so an advanced-search submission with active locks doesn't auto-open the breadcrumb the way #2332 intends. Needs the matching `import { markJustSubmitted } from '@/composables/useSearchFilter'` at the top too.

```js
function handleAdvancedSearch({ query, field }) {
  searchStore.setQuery(query)
  searchStore.setField(field)
  markJustSubmitted()
  // Resets `from` and stamps the route, so an unchanged query still resubmits.
  refreshRouteFromStart()
}
```

(File-level because `handleAdvancedSearch` is outside this PR's hunks.)

src/store/modules/search.js
```diff
@@ -886,41 +915,112 @@ export const useSearchStore = defineSuffixedStore('search', () => {
...
+    const groupExcluded = new Map()
+    lockedFiltersStore.entries.forEach((entry) => {
+      const { bareName, excluded, exists } = getLockConflict(entry)
+      if (!exists) {
+        return
+      }
+      const canonical = getCanonicalDimension(bareName)
+      if (canonical === bareName || !groupExcluded.has(canonical)) {
+        groupExcluded.set(canonical, excluded)
+      }
```
pirhoo, line 1058: important: for an unpaired filter `canonical === bareName` always holds, so this guard never protects anything and the last matching lock entry wins. With `language`/ENGLISH locked included and `-language`/FRENCH locked excluded (reachable via a shared exclude link plus locking a second value), Apply leaves `language` excluded, silently inverting the include-lock, and since `hasConflictingLocks` stays true the footer's `v-else-if` permanently hides "Unlock filters (N)". Restricting the override to a genuine pair keeps the canonical precedence you documented and makes the unpaired case first-wins instead of order-dependent.

```suggestion
      const canonical = getCanonicalDimension(bareName)
      const isCanonicalOfPair = canonical === bareName && getPairedDimensions(bareName).length > 1
      if (isCanonicalOfPair || !groupExcluded.has(canonical)) {
        groupExcluded.set(canonical, excluded)
      }
```

src/store/modules/search.js
```diff
@@ -886,41 +915,112 @@ export const useSearchStore = defineSuffixedStore('search', () => {
...
+      if (!exists) {
         return
       }
       addFilterValue({ name: bareName, value })
-      if (excluded) {
+      if (groupExcluded.get(getCanonicalDimension(bareName))) {
         excludeFilter(bareName)
       }
+      else {
+        includeFilter(bareName)
```
pirhoo: important: `includeFilter` here only clears the bare dimension, then the trailing `reconcilePairedExcludeFilters()` re-excludes it from the still-excluded sibling. With `?f[-contentType]=application/pdf` (which hydrates both dimensions as excluded) plus an include-mode lock on `contentType`, `excludeFilters` still ends up `['contentTypeCategory', 'contentType']` after Apply, so `hasConflictingLocks` never clears. Could we clear the whole group? `getPairedDimensions` already returns `[name]` when unpaired, so it's safe either way.

```suggestion
        getPairedDimensions(bareName).forEach(includeFilter)
```

src/composables/useSearchBreadcrumb.js
```diff
@@ -220,6 +225,17 @@ export function useSearchBreadcrumb() {
...
+  const applyLockedFilters = async () => {
+    try {
+      searchStore.applyLockedFilters()
+      await refreshRouteFromStart()
+      toast.success(t('searchBreadcrumbFormFooter.applyLockedFiltersSuccess'))
+    }
+    catch {
+      toast.error(t('searchBreadcrumbFormFooter.applyLockedFiltersError'))
+    }
+  }
```
pirhoo, line 233: suggestion: the success toast is unconditional, and since the store call is synchronous and `router.push` also resolves on an aborted navigation, nothing here can actually fail into the `catch`. Checking that the conflict actually cleared would make the toast mean something, and would surface the two cases above instead of reporting success over them.

```suggestion
  const applyLockedFilters = async () => {
    try {
      searchStore.applyLockedFilters()
      await refreshRouteFromStart()
      if (searchStore.hasConflictingLocks) {
        toast.error(t('searchBreadcrumbFormFooter.applyLockedFiltersError'))
        return
      }
      toast.success(t('searchBreadcrumbFormFooter.applyLockedFiltersSuccess'))
    }
    catch {
      toast.error(t('searchBreadcrumbFormFooter.applyLockedFiltersError'))
    }
  }
```

pirhoo (changes requested):
Noting the first one here rather than inline, since `src/router/guards/prefillSearchStore.js` is not part of this diff.

**`src/router/guards/prefillSearchStore.js`**

crucial: this guard runs as part of the `search` route's `beforeEnter` (see `src/router/index.js`), so it calls `updateFromRouteQuery` with `mergeLocks` defaulting to `true` before Search.vue ever gets to read `isSavedSearchOpened()`. That looks like it defeats the whole #2331 skip: opening a saved search still injects every active lock. It's outside the diff, which is probably why it slipped.

```js
import { isSavedSearchOpened } from '@/composables/useSearchFilter'
import { useSearchStore } from '@/store/modules'

export const prefillSearchStore = (to) => {
  const searchStore = useSearchStore()
  searchStore.updateFromRouteQuery(to.query, { mergeLocks: !isSavedSearchOpened() })
}
```

src/views/Search/Search.vue
caro3801: Done in 5f9baed70. Submitting the advanced search modal is meant to auto-open the breadcrumb when locks are active, the same way a regular search bar submit does, so it needed the same one-shot flag SearchBar.vue sets before its own refresh.

src/store/modules/search.js
caro3801: Done in 1ca7a8775, with your suggested diff. The canonical-precedence guard is meant to make the pair's own canonical lock win over its partner, but for an unpaired filter it always evaluated true, so the safeguard never engaged there and the group just resolved to whichever entry got processed last.

src/store/modules/search.js
caro3801: Done in a062a5518, with your suggested diff. An include-mode lock is meant to actually clear the conflict it resolves, so it needs to release both sides of a paired group, not just the bare dimension the lock names.

src/composables/useSearchBreadcrumb.js
caro3801: Done in 5a2628811, with your suggested diff. The toast is meant to report whether the conflict actually cleared, not just whether the call ran without throwing, so it now checks hasConflictingLocks after the route refresh.

pirhoo (approved):
Although the build is heavily broken, it's approved, thanks for the changes!

caro3801:
Yes, a big part was because of a bad import in the tests/unit/setup.js . It pulled useLockedFiltersStore from the @/store/modules , whose first export re-exports search.js. Barrel imports evaluated the whole file, so this eagerly loaded search.js -> apiInstance -> elasticsearch -> elasticsearchClient -> real axios, once per worker,
before any spec file's own vi.mock('axios') could hoist and intercept it.

That silently defeated axios mocking suite-wide: elasticsearchClient.spec.js, api/index.spec.js,  atchSearchActions.spec.js and others started issuing real HTTP requests and failing with Network Error.

Fixed in [28ffc24](https://github.com/ICIJ/datashare-client/pull/503/commits/28ffc24588b00afe3c094bfaf890cc88a7b4b68c) by importing from @/store/modules/lockedFilters directly instead.

PR #504, opened 2026-08-21
feat(search): extend lock support to remaining filter types

caro3801:
## Summary

Extends lock support (icij/datashare#2219) to every filter type the original 6-story epic skipped — anything that overrides `FilterType.vue`'s default slot instead of using it: `recommendedBy`, `path`, and content-type categories (locking a category like "Document" as one bulk unit, separate from locking individual types inside it). Fixes several bugs found while building and reviewing that support, most from live browser testing rather than unit tests.

`project` stays deliberately unlockable — locking it would fight the entire point of switching projects (product decision, not deferred).

## Decisions

- **Path filter has no per-value row to hang a lock icon on**, unlike every other filter type here. After 4 design iterations (bulk-lock action → lock-only-on-selected → select+lock hover-revealed → final: lock icon visible on *every* row, hover-revealed, select+lock on click), landed on: lock button shows on hover for any row, clicking an unlocked row selects *and* locks it in one action, matching the select+lock convention every other type in this PR uses.
- **Document-type categories lock as one bulk unit** (`contentTypeCategory=DOCUMENT`), not as N individual `contentType` locks — mirrors how the category is already stored as a single filter value.
- **Locking a category-covered leaf value demotes the category** instead of stacking underneath it: locking e.g. "PDF" while "Document" is locked now releases the Document lock/value and selects+locks PDF explicitly, rather than silently doing nothing but light up the lock icon. (Found via manual testing after the initial implementation shipped the opposite, intentionally-conflict-avoiding behavior — see Changes.)
- **Path filter no longer resets its selection when a project is removed.** ES already scopes each query to the selected projects, so a path left over from a removed project just stops matching anything instead of producing wrong results — no client-side reset needed, and no more wiping unrelated valid selections when only one of several selected projects drops out (discussed in #504's own review thread).
- **`isFilterExcluded` now uses canonical-dimension precedence**, not OR-across-both-sides, for a paired dimension's exclude-mode read — matching the precedence rule `getCanonicalDimension` already documents and `applyLockedFilters` already follows, so a stale exclude on the non-canonical side alone no longer flips the pair's read.

## Changes

- **`recommendedBy`**: lock/unlock wired through the existing bucket-based pattern (bare lock name, no exclude mode).
- **`path`**: select+lock on any row (hover-revealed lock icon), plus three bugs found and fixed along the way — two click-through regressions where a new interactive element (caret, lock button) sat under `PathTreeViewEntryName`'s stretched-link overlay and silently ate clicks, and a duplicate-breadcrumb-chip bug from locking against an unnormalized path value.
- **Document-type categories**: category-level lock/unlock, select+lock on an unselected category, and (this session) fixed locking a category-covered leaf value to demote the category instead of stacking under it.
- **Lock icon color/style**: switched to the filled Phosphor variant (`lock-fill`/`lock-open-fill`) everywhere; fixed the locked-state color being hardcoded to a raw token that never adapted for dark mode (unreadable navy-on-dark), and the unlocked icon's hover color falling back to a near-black default instead of the intended tertiary tone.
- **`FilterTypePath`**: dropped the auto-reset-on-project-removal behavior (see Decisions).
- **`isFilterExcluded`** (`useSearchFilter.js`): switched from OR-across-both-sides to canonical-dimension precedence for a paired dimension's exclude-mode read (see Decisions).
- Assorted fixes surfaced by review/testing: exclude-filter mode not reflected in `FilterTypeFileTypes` locks, a lock's dimension not retagging to the mode it's (re)locked in, missing locked filters not rendering until real ES buckets existed, hideLock support gaps in `FilterTypePath`/`FilterTypeRecommendedBy`, and a couple of stale/incorrect tests (one asserting `recommendedBy` never wires up locking — it does, as of this PR; one building an impossible same-dimension double-lock scenario).

## Important files

- `src/composables/useContentTypeSelection.js` — category ↔ leaf-type promote/demote logic; the demotion path (`toggleEntry` → `handleEntryCheck` → `checkChildOfStoredCategory`) is what the category-lock-demotion fix now routes through.
- `src/components/Filter/FilterType/FilterTypeFileTypes.vue` — leaf + category lock wiring (`toggleLock`, `toggleLockCategory`).
- `src/components/Filter/FilterType/FilterTypePath.vue` / `src/components/PathTree/PathTreeView/PathTreeViewEntry.vue` — path's select+lock wiring and the stretched-link click-through fixes.
- `src/components/Button/ButtonToggleLock.vue` — single shared lock-icon component (icon choice, locked-state and hover colors) used by every filter type in this PR plus the breadcrumb.
- `src/store/modules/search.js` — `isFilterExcluded`, `getLockConflict`/paired-dimension conflict logic.
- `src/store/modules/lockedFilters.js` — the lock store itself, including the retag-to-current-mode invariant that same-dimension opposite-mode locks now rely on.


caro3801:
**Question: how should path selection behave when a project is removed?**

`FilterTypePath.vue`'s `watchIndices` reset previously fired on *any* project change, including adding one — which also silently unlocked any locked path filters. I scoped it down to only reset when a project actually drops out of the current selection (21a1756f2), but that still over-resets: it wipes the **entire** path selection on any removal, not just the removed project's paths.

The reason it can't be scoped further: a selected path is a plain directory string with no project tag, and `dataDir` is a single global config value — there's no client-side way to tell "this path came from `cantina`" vs "this path came from `local-datashare`".

Concretely: select a path under `local-datashare`, add `cantina`, then remove `cantina` — the `local-datashare` path gets cleared too, even though it's still valid.

Two ways to go:
1. **Drop the auto-reset entirely.** ES already scopes each query to the selected projects, so a stale path from a removed project just stops matching anything — no incorrect results, it just silently stops contributing. Matches how this epic already tolerates locked-but-not-in-current-buckets entries elsewhere.
2. **Keep resetting on any removal** (current behavior) and accept that removing one project among several can clear unrelated valid selections/locks.

Leaning towards (1) - wanted the team's take before changing it, since it's a product-behavior call, not just a bug fix.

current behavior
[Screencast from 2026-08-25 19-36-16.webm](https://github.com/user-attachments/assets/8ef0079d-5510-4f23-aea3-cf45724fd943)



edit: Solution chosen is 1. implemented in [d5bccf3](https://github.com/ICIJ/datashare-client/pull/504/commits/d5bccf3cff27855dd41a4bd6b9ed0ced9c6924b2)

src/composables/useContentTypeSelection.js
```diff
@@ -34,6 +34,15 @@ export function useContentTypeSelection({ filter, categories }) {
...
   }
 
+  // Same bypass-of-useSearchFilter reasoning as unlockContentType, for the
+  // category's own lock (a separate lock namespace under
+  // CONTENT_TYPE_CATEGORY_FILTER_NAME, since the category is stored as its
+  // own bulk value rather than as N individual contentType entries).
+  const unlockCategory = (category) => {
+    const name = toLockedName(CONTENT_TYPE_CATEGORY_FILTER_NAME, isFilterExcluded({ name: CONTENT_TYPE_CATEGORY_FILTER_NAME }))
+    lockedFiltersStore.unlock({ name, value: category })
+  }
```
pirhoo: crucial: `promoteToCategory` below unlocks every type the category covers but never locks the category in its place, so locking the last unlocked type of a category silently destroys the sibling locks (persisted) and leaves the just-locked type as an orphan lock for a value the filter no longer stores explicitly. Could we transfer the lock instead? That needs an `isContentTypeLocked` / `lockCategory` pair next to these two helpers, then:

```js
const promoteToCategory = (category) => {
  const categoryTypes = typesInCategory(category)
  const wasLocked = categoryTypes.some(isContentTypeLocked)
  categoryTypes.forEach(unlockContentType)
  if (wasLocked) {
    lockCategory(category)
  }
  writeContentTypes(currentContentTypes().filter(value => !categoryTypes.includes(value)))
  writeCategories([...currentCategories(), category])
}
```

src/components/Filter/FilterType/FilterTypePath.vue
```diff
@@ -31,16 +33,78 @@ const props = defineProps({
   },
   hideCount: {
     type: Boolean
+  },
+  // Suppresses the per-row lock button entirely. See FilterType.vue's
+  // hideLock for why (disposable/unrelated screens rendering against a
+  // non-live search store).
+  hideLock: {
+    type: Boolean
   }
```
pirhoo, line 42: important: looks like `hideLock` is accepted here but never forwarded to the `filter-type` in the template, so `FilterModal` re-renders this with the default `hide-lock=false` and expanding the path filter in the batch-search form gives lock buttons that write to the user's real persisted lock store. Adding `:hide-lock="hideLock"` to that `<filter-type>` should cover it.

src/components/Filter/FilterType/FilterTypeRecommendedBy.vue
```diff
@@ -79,6 +119,9 @@ watchIndices(fetch)
           :value="user"
           :count="count"
           :hide-count="hideCount"
+          :lockable="!hideLock"
+          :locked="isItemLocked(user)"
+          @update:locked="toggleLock(user, $event)"
```
pirhoo, line 126: important: the rows honour `hideLock`, but the wrapping `<filter-type>` above never receives it, so `FilterTypeAll`'s "All" toggle can still unlock real `recommendedBy` locks from the batch-search form. Same one-line fix as the path filter: `:hide-lock="hideLock"` on that `<filter-type>`.

src/views/Search/Search.vue
```diff
@@ -156,19 +156,44 @@ onAfterRouteQueryFromUpdate(refreshSearchFromRoute, { immediate: route.name ===
...
+stopConflictWatch = watch(
+  hasConflictingLocks,
+  (conflict) => {
+    if (conflict) {
+      toggleSearchBreadcrumb.value = true
+      stopConflictWatch?.()
+    }
+  },
+  { immediate: true }
+)
```
pirhoo: important: on the `immediate` run `stopConflictWatch` is still null, so the `?.()` no-ops and the "open once" watcher is never stopped in exactly the shared-link case it was written for; a later conflict then re-opens the panel the user closed. Latching a flag covers that first run:

```suggestion
// Initialized to null (rather than assigned in the same statement as the
// watch() call below) so the immediate callback below — which runs
// synchronously, before that assignment completes — reads an already
// -initialized null instead of hitting the temporal dead zone. The flag
// covers that same run: the handle isn't there yet, so "open once" has to
// be latched rather than relying on stopping the watcher.
let conflictHandled = false
let stopConflictWatch = null
stopConflictWatch = watch(
  hasConflictingLocks,
  (conflict) => {
    if (!conflict || conflictHandled) {
      return
    }
    conflictHandled = true
    toggleSearchBreadcrumb.value = true
    stopConflictWatch?.()
  },
  { immediate: true }
)
```

src/components/ContentTypes/ContentTypesCategories/ContentTypesCategoryName.vue
```diff
@@ -58,6 +64,9 @@ const collapseLabel = computed(() => t('contentTypesCategoryName.toggle', { cate
       :label="resolvedLabel"
       :count="count"
       :indeterminate="indeterminate"
+      :locked="locked"
+      lockable
+      @update:locked="emit('update:locked', $event)"
     />
```
pirhoo, line 73: important: `lockable` is hardcoded and there's no `hideLock` prop, so with the grouped view on the category lock buttons still render (and write to the persisted lock store) inside the batch-search content-type modal. Could we add a `hideLock` boolean prop alongside `locked` and pass `:hide-lock="hideLock"` from the `content-types-category-name` call site in `FilterTypeFileTypes.vue`?

```suggestion
      :label="resolvedLabel"
      :count="count"
      :indeterminate="indeterminate"
      :locked="locked"
      :lockable="!hideLock"
      @update:locked="emit('update:locked', $event)"
    />
```

src/components/PathTree/PathTreeView/PathTreeViewEntry.vue
```diff
@@ -373,5 +395,26 @@ const hasChildren = computed(() => !!slots.default)
...
+  // Same hidden-until-hover/focus/locked idiom as the Filters panel row's own
+  // lock button — ButtonToggleLock itself only owns the locked-state color.
+  &__lock {
+    opacity: 0;
+    transition: opacity 0.15s ease;
+
+    &.button-toggle-lock--locked {
+      opacity: 1;
+    }
+  }
```
pirhoo, line 410: important: this drops the `@media (hover: hover)` guard the sibling row in `FiltersPanelSectionFilterEntry.vue` uses, so on touch the lock button becomes an invisible but tappable toggle (nothing ever matches `--active` there to reveal it).

```suggestion
  &__lock {
    transition: opacity 0.15s ease;
    @media (hover: hover) {
      opacity: 0;
    }

    &.button-toggle-lock--locked {
      opacity: 1;
    }
  }
```

src/store/modules/lockedFilters.js
```diff
@@ -114,6 +125,25 @@ export const useLockedFiltersStore = defineStore('lockedFilters', () => {
...
+   * @param {string} params.name - The entry's current name.
+   * @param {string} params.newName - The name to rename it to.
+   * @param {string|number} params.value - The filter value.
+   */
+  function retag({ name, newName, value }) {
+    const index = indexByKey.value.get(entryKey(name, value))
+    if (index !== undefined) {
+      entries.value[index] = { ...entries.value[index], name: newName }
+    }
+  }
```
pirhoo, line 201: suggestion: `retag` doesn't dedupe against an existing `{ newName, value }`, so it can leave two entries sharing one key; `indexByKey` maps only the last, and one unlock click would leave the lock apparently still set. `lock`'s own retag pass keeps the two modes from coexisting today, so this is latent rather than reachable, but the guard is cheap:

```suggestion
  function retag({ name, newName, value }) {
    const index = indexByKey.value.get(entryKey(name, value))
    if (index === undefined) {
      return
    }
    const alreadyTagged = indexByKey.value.has(entryKey(newName, value))
    if (alreadyTagged) {
      entries.value.splice(index, 1)
      return
    }
    entries.value[index] = { ...entries.value[index], name: newName }
  }
```

pirhoo (changes requested):

src/composables/useContentTypeSelection.js
pirhoo: Half of this landed, thanks — the `update:locked` path now locks the category instead of orphaning the leaf, and the new spec pins it.

The plain-checkbox path is still open though: `promoteToCategory` is unchanged, and `@update:model-value="toggleEntry(contentType, $event)"` in `FilterTypeFileTypes.vue` routes straight through it, so the sibling locks are dropped with nothing put in their place. Adding this to your own "unlocks a locked child when it gets promoted into its category by ticking the last sibling" spec fails on the current head:

```js
expect(lockedFiltersStore.isLocked({ name: 'contentTypeCategory', value: 'OTHER' })).toBe(true)
// AssertionError: expected false to be true
```

So whether a persisted lock survives completing a category now depends on whether the user finished with a lock click or a plain tick. Could the transfer live in `promoteToCategory` itself, so both entry points inherit it?

src/components/Filter/FilterType/FilterTypePath.vue
pirhoo: Fixed, thanks.

src/components/Filter/FilterType/FilterTypeRecommendedBy.vue
pirhoo: Fixed, thanks.

src/views/Search/Search.vue
pirhoo: Resolved — dropping the self-stop (and the now-stale null-init comment) is cleaner than latching a flag, and the new "second, unrelated conflict" spec pins the intended behaviour.

src/composables/useContentTypeSelection.js
caro3801: Fixed in 500b13fb4 - the transfer now lives in promoteToCategory itself, so both the lock-click and plain-tick paths inherit it.

src/components/ContentTypes/ContentTypesCategories/ContentTypesCategoryName.vue
caro3801: Fixed in dcdde514a - added the `hideLock` prop to `ContentTypesCategoryName.vue`, wired `:lockable="!hideLock"`, and forwarded it from the `FilterTypeFileTypes.vue` call site.

src/components/PathTree/PathTreeView/PathTreeViewEntry.vue
caro3801: Added (again ^^") in 987f45e20,  restored the `@media (hover: hover)` guard so touch never gets an invisible tappable lock button.

src/store/modules/lockedFilters.js
caro3801: Fixed in 446e978e0, using your suggested diff - retag now drops the stale entry instead of leaving a same-key duplicate.

caro3801:
Thanks for the review and the suggestions

While verifying the reviewed fixes, found and closed two more `hideLock` leaks in the same family as the ones above - both let a disposable/non-live render (e.g. the batch-search creation form) write to the user's real, persistent lock store:

- 1f6039c0b - grouped content-type unticking/promotion (`useContentTypeSelection.js`) never received `hideLock`, so it unlocked through regardless.
- d2d33adbb - `FilterTypePath.vue`/`FilterTypeRecommendedBy.vue`'s `unlockRemovedValues` call ran unconditionally on deselect, same gap.

Also a couple of small cleanups surfaced along the way:

- 8731bd2a0 / aac41d2a0 - merged a duplicated `hideLock` guard and deduped a lock-lookup computed that `useContentTypeSelection.js` already provided.
- f709f6c4a - `missingLocks` (`FilterType.vue`) was scanning the *entire* shared lock store on every recompute, in every one of the ~10-15 mounted filter facets; added `entriesForName` to `lockedFiltersStore` so each facet does an O(1) lookup instead.

pirhoo (approved):
Looks good, thanks!

PR #509, opened 2026-08-26
fix(search): Apply locked filters UX polish icij/datashare#2332

caro3801:
Stacked on #504.

Follow-up polish to icij/datashare#2332's "Apply locked filters" action:

- Reverses the earlier "silent-merge-unless-conflicting" design: a locked
  filter absent from the URL/route no longer auto-applies. "Apply locked
  filters" now shows whenever a locked filter isn't reflected in the live
  search, and only clicking it applies those values. "Clear filters" still
  reapplies locks instantly (that's already an explicit user action).
- "Apply locked filters" and "Clear locks" (renamed from "Unlock filters")
  are now always visible, side by side, and only toggle between
  enabled/disabled instead of popping in/out of the DOM — avoids layout
  jank next to each other on every navigation.
- Disabled-state tooltip on "Apply locked filters", split by reason
  ("All locked filters are already applied" vs "No locks to apply"), via a
  native `title` on a wrapping span rather than the button itself, since a
  disabled native `<button>` never fires the mouse events a JS tooltip
  needs.
- Lock-count badge (replacing the old "(N)" text) colored to match the
  active-filters badge in the filters panel, including on hover.

[Screencast from 2026-08-26 19-20-55.webm](https://github.com/user-attachments/assets/aeed0795-b389-4193-ba22-ad8cbcdbf442)


src/views/Search/Search.vue
```diff
@@ -154,10 +153,6 @@ onAfterRouteQueryFromUpdate(refreshSearchFromRoute, { immediate: route.name ===
 // refresh guards above so it runs last and they observe the flag before it is
 // stripped from the URL.
 onConsumeNoRefresh({ immediate: route.name === 'search' })
-// Consume the one-shot `savedSearchOpened` flag. MUST be registered after
-// the two refresh guards above so it runs last and both observe the flag
-// before it is cleared here.
-onConsumeSavedSearchOpened({ immediate: route.name === 'search' })
 // Auto-open the breadcrumb panel after an explicit search submission when
 // locks are active, so the user immediately sees what's locked. Gated on a
 // one-shot in-memory flag (set only by SearchBar's submit(), consumed here)
```
pirhoo, line 158: important: about the conflict watcher just below — could we hold it off until hydration? On a cold load it runs during setup, before the route is applied, so `hasConflictingLocks` is still true for any persisted lock and the panel opens even when the URL already carries that exact value. Gating on a real mode conflict (`isExcluded !== excluded`) would do it.

src/components/Search/SearchParameter/SearchParameterQueryTerm.vue
```diff
@@ -103,8 +111,8 @@ const showOperator = computed(() => {
     :icon-left-label="iconLabel"
     :icon-right="noXIcon ? null : IPhX"
     icon-right-hover-weight="bold"
-    tag="span"
-    role="presentation"
+    :tag="isLockable ? 'span' : undefined"
+    :role="isLockable ? 'presentation' : undefined"
```
pirhoo: important: `locked` defaults to null and `queryComponentProps` never forwards it, so `isLockable` is false for every query chip. Read-only breadcrumbs in saved searches and batch search end up as focusable buttons with nothing behind them. Gating on the click handler keeps the `DocumentGlobalSearchTermsEntry` case working:

```suggestion
    :tag="$attrs.onClick ? undefined : 'span'"
    :role="$attrs.onClick ? undefined : 'presentation'"
```

src/components/Search/SearchBreadcrumbForm/SearchBreadcrumbFormFooter.vue
```diff
@@ -49,23 +65,38 @@ const emit = defineEmits(['clear:filters', 'clear:query', 'clear:all', 'unlock:a
...
-        {{ t('searchBreadcrumbFormFooter.applyLockedFilters') }}
-      </button-icon>
+        <button-icon
+          :disabled="!hasConflictingLocks"
+          :icon-left="IPhLockOpen"
+          @click="emit('apply:locked-filters')"
+        >
+          {{ t('searchBreadcrumbFormFooter.applyLockedFilters') }}
+        </button-icon>
+      </span>
```
pirhoo, line 96: important: the wrapper span makes the button `li > span > .btn`, so `FormActionsCompactDropdown`'s `> li > .btn` rules stop matching below the `md` breakpoint and it loses its full-width and dropdown-link styling. Could we put `:title` on the `button-icon` itself and drop the span? That also gets keyboard and screen reader users the explanation, which a title on a span wrapping a disabled button never reaches.

src/core/FiltersMixin.js
```diff
@@ -79,7 +80,10 @@ const FiltersMixin = superclass =>
       return this.toggleForProject({
         project,
         // Conditional callbacks
-        withFn: () => this.unregisterFilter(name),
+        // preserveLocks: this filter is only hidden for this project
+        // selection, not permanently removed - registerFilter below never
+        // restores locks, so purging them here would lose them for good.
+        withFn: () => this.unregisterFilter(name, { preserveLocks: true }),
```
pirhoo, line 89: important: nice fix, but `registerFilterForProject`'s `withoutFn` above (line 65) is the same case and still purges the locks. A filter registered only for project A loses them for good on switching to project B, even though it is only hidden. Same one-word fix there: `this.unregisterFilter(options.name, { preserveLocks: true })`.

src/components/Search/SearchParameter/SearchParameterQueryAst.vue
```diff
@@ -115,6 +116,7 @@ const effectiveRightOperator = computed(() => {
       :no-x-icon="noXIcon"
       :size="size"
       @click:x="emit('click:x', $event)"
+      @click:lock="emit('click:lock', $event)"
```
pirhoo, line 121: suggestion: the left recursion at the top of the template did not get this, or `v-bind="$attrs"`. Nothing forwards `locked` down this path yet, but lucene nests left-associatively, so `a AND b AND c` puts the common case on exactly the branch that drops it.

pirhoo (changes requested):
Almost there!

src/views/Search/Search.vue
caro3801: Traced the actual mount order with a temporary stack-trace log rather than reasoning about it: the watch does run
during setup, before Search's own onMounted hooks, but the 'search' route's own beforeEnter guard
(prefillSearchStore, src/router/index.js:74) already hydrates the store from `to.query` before Search ever mounts.
By the time the watch evaluates, the store already reflects the URL, so hasConflictingLocks never actually sees a
pre-hydration state in practice. Added a regression test on a genuinely fresh store (c942838c2) pinning this, since
the existing on-mount tests reuse an already-hydrated store across a wrapper.unmount()+remount within the same test
and don't exercise a real cold boundary. Let me know if you had a different repro path in mind that doesn't go
through that guard.

src/components/Search/SearchParameter/SearchParameterQueryTerm.vue
caro3801: Done in 139bc96fb. A read-only breadcrumb display with no click listener attached no longer renders as a focusable
button with nothing behind it to activate; DocumentGlobalSearchTermsEntry's clickable term chips keep their native
button, since those do attach one.

src/components/Search/SearchBreadcrumbForm/SearchBreadcrumbFormFooter.vue
caro3801: Kept the wrapping span rather than dropping it: it's there because a disabled native button has
pointer-events:none, which also blocks the native title tooltip from firing on hover, not just bootstrap-vue-next's
v-b-tooltip directive. Dropping it would silence the mouse-hover tooltip on the disabled state again. Fixed the
underlying CSS breakage instead (ceb68e42d): FormActionsCompactDropdown's `> li > .btn` selector required the
button to be a direct child of the `<li>`, which the wrapper broke below the md breakpoint; widened it to a
descendant selector. Also put the same title directly on the button itself, so a screen reader or keyboard-focus
user gets it too, not just a sighted mouse user hovering the wrapper.

src/core/FiltersMixin.js
caro3801: Done in 0bce59879, same fix applied to registerFilterForProject's withoutFn. Also fixed removeFilter's `delete
filters.value[i]`, found while adding the regression test: it left an undefined hole in the array instead of
removing the element, crashing the next addFilter's own .find() over the same array as soon as a test exercised a
registration after a removal in the same file.

src/components/Search/SearchParameter/SearchParameterQueryAst.vue
caro3801: Done in c6d30d3c9. The left branch now forwards $attrs and re-emits click:lock the same way the right one does, with
a regression test covering the left-associative nesting case directly.

caro3801:
Also did a self-review pass beyond the inline comments above; posting a summary here rather than as
individual line comments since none of it was flagged by review:

- **`SearchBar.vue`** (8ed274fce): `submit()` was calling the lock-unaware `resetFilterValues()`
  instead of `resetFilterValuesPreservingLocks()` on a clear-filters submit, leaving stale
  `excludeFilters` behind and skipping a force-apply of locks.
- **`FilterDate.js` / `FilterDateRange.js`** (ee5fd4e2f): locking the `indexingDate` or
  `creationDate` chip rendered an undefined or raw unformatted label. Both `itemLabel()`s now format
  a route-query string value (including `creationDate`'s `"min:max"` pair), and both guard against a
  null `localStorage` locale, a separate crash this fix's own tests turned up (dayjs's `.locale(x)`
  treats a falsy `x` as a getter, silently breaking `.format()`).
- **`useSearchBreadcrumb.js`** (3de480218): the `applyLockedFilters` catch block swallowed every
  error behind a generic toast, indistinguishable from an expected conflict in a bug report. Now logs
  before toasting.
- **`search.js` / `FiltersMixin.js`** (e3f3ceebb, ahead of the fix above for your `FiltersMixin.js:86`
  comment): `removeFilter` unconditionally purged a filter's locks, reachable through
  `unregisterFilterForProject`, which only hides a project-scoped filter and never restores locks on
  re-registration, so a project switch lost them for good. Added a `preserveLocks` option, set by
  `unregisterFilterForProject` only; a bare `unregisterFilter`/`useSearchFilter`'s `removeFilter` still
  purge as before.
- **`search.js`** (3f39efcdf, doc only): `applyLockedFilters` resolves one include/exclude mode per
  filter and applies it to the whole filter, so it can flip a co-resident unlocked value's mode too
  (e.g. a locked `FRENCH` exclude pulls an unlocked `ENGLISH` include along with it). Not a one-line
  fix given the single-mode-per-filter data model; confirmed with product this is intentional (locks
  win for the whole filter), added a doc comment plus a regression test pinning it down.
- Also folded into `0bce59879` while adding its regression test: `FiltersMixin.js`'s `removeFilter`
  was doing `delete filters.value[i]`, leaving an undefined hole in the array instead of removing the
  element, which crashed the next `addFilter`'s `.find()` over the same array.

Unrelated to review, 6568af1d9 and b868bf6a6 (unlock-both-modes-on-clear regression, em dash
cleanup) also landed on this branch during the rebase; noting them here for completeness since they
weren't split into their own PR.

pirhoo (approved):
Super!

src/views/Search/Search.vue
caro3801: Traced the actual mount order with a temporary stack-trace log rather than reasoning about it: the watch does run during setup, before Search's own onMounted hooks, but the 'search' route's own beforeEnter guard (prefillSearchStore, src/router/index.js:74) already hydrates the store from `to.query` before Search ever mounts. By the time the watch evaluates, the store already reflects the URL, so hasConflictingLocks never actually sees a pre-hydration state in practice. Added a regression test on a genuinely fresh store (c942838c2) pinning this, since the existing on-mount tests reuse an already-hydrated store across a wrapper.unmount()+remount within the same test and don't exercise a real cold boundary. Let me know if you had a different repro path in mind that doesn't go through that guard.

src/components/Search/SearchParameter/SearchParameterQueryTerm.vue
caro3801: Done in 139bc96fb. A read-only breadcrumb display with no click listener attached no longer renders as a focusable button with nothing behind it to activate; DocumentGlobalSearchTermsEntry's clickable term chips keep their native button, since those do attach one.

src/components/Search/SearchBreadcrumbForm/SearchBreadcrumbFormFooter.vue
caro3801: Kept the wrapping span rather than dropping it: it's there because a disabled native button has pointer-events:none, which also blocks the native title tooltip from firing on hover, not just bootstrap-vue-next's v-b-tooltip directive. Dropping it would silence the mouse-hover tooltip on the disabled state again. Fixed the underlying CSS breakage instead (ceb68e42d): FormActionsCompactDropdown's `> li > .btn` selector required the button to be a direct child of the `<li>`, which the wrapper broke below the md breakpoint; widened it to a descendant selector. Also put the same title directly on the button itself, so a screen reader or keyboard-focus user gets it too, not just a sighted mouse user hovering the wrapper.

src/core/FiltersMixin.js
caro3801: Done in 0bce59879, same fix applied to registerFilterForProject's withoutFn. Also fixed removeFilter's `delete filters.value[i]`, found while adding the regression test: it left an undefined hole in the array instead of removing the element, crashing the next addFilter's own .find() over the same array as soon as a test exercised a registration after a removal in the same file.

src/components/Search/SearchParameter/SearchParameterQueryAst.vue
caro3801: Done in c6d30d3c9. The left branch now forwards $attrs and re-emits click:lock the same way the right one does, with a regression test covering the left-associative nesting case directly.

PR #510, opened 2026-08-28
fix: false "Unsupported format" warning on audio and video player

mvanzalu:
## Summary

The audio and video players used canPlayType() to check if the browser supports the file format. This method tests the MIME type string assigned by Tika, which often differs from the standard types browsers recognize (for example audio/vnd.wave instead of audio/wav, audio/vorbis instead of audio/ogg). This caused false "Unsupported format" warnings on files the browser could play perfectly fine. 

## Proposal

The fix replaces this with actual player events: @error with MEDIA_ERR_SRC_NOT_SUPPORTED fires only when the browser truly cannot decode the file, and @canplay clears the warning when playback succeeds.


pirhoo (approved):
Thanks!

PR #511, opened 2026-08-28
fix(path-tree): selected/hovered row colors, murmur/BVN bump, no auto-expand on select

caro3801:
Bumps `@icij/murmur` to 5.1.3 and `bootstrap-vue-next` to 1.0.1 (fixes toggle buttons staying visually pressed after click, icij/datashare#2332), and fixes path-tree row styling and behavior:

- icij/datashare#2365: selected/hovered path rows no longer get the wrong background/text colors (was a full navy fill with orphaned white text on some rows, wrong badge color/guard on others).

- fixed also the path filter that was wrongly showing a magnifying glass icon on hover

- Ticking a directory in the path filter no longer auto-expands it - only the caret or the name does.

image

image

clear filters hover
image

after clear locks, no more primary button in the bottom of the breadcrumb
image


src/components/PathTree/PathTreeView/PathTreeViewEntry.vue
```diff
@@ -376,12 +376,6 @@ function toggleLock(value) {
     --path-tree-view-entry-header-bg: transparent;
   }
```
pirhoo, line 378: important: removing the `&--selected:not(&--compact)` header rule also drops the selected-row highlight in the folder picker, where `FormControlPath` passes `no-stats` so the navy document count that replaces it never renders (ProjectForm, TaskDocumentsForm, PathBannerForm). Could we keep a subtle fill so the affordance survives without the white-on-navy problem?

```suggestion
  }

  &--selected:not(&--compact) > &__header {
    --path-tree-view-entry-header-bg: var(--bs-tertiary-bg-subtle);
  }

```

src/components/PathTree/PathTreeView/PathTreeViewEntry.vue
```diff
@@ -264,7 +264,7 @@ function toggleLock(value) {
           :directories="directories"
           :size="size"
           :selected="selected"
-          :active="compactOrInjected ? selected : active"
+          :active="compactOrInjected ? selected : active || selected"
```
pirhoo: important: `active || selected` means hover paints the count navy too, and with the `:not(&--no-search-link)` guard gone that now hits the path filter as well, where the count is a plain `span`: a navy pill that looks clickable but isn't, and hovered rows look selected. Gating on `noSearchLink` (already a prop here) would keep it to selection only.

```suggestion
          :active="compactOrInjected || noSearchLink ? selected : active || selected"
```

src/components/Filter/FilterType/FilterTypePath.vue
```diff
@@ -104,12 +104,16 @@ provide('pathLockable', pathLockable)
...
 provide('toggleLockPath', toggleLockPath)
 
-// Pre-open ancestor directories of selected paths so the tree
-// reveals them immediately instead of requiring manual expansion.
+// Pre-open ancestor directories of selected paths so the tree reveals them
+// immediately instead of requiring manual expansion. getAncestorPaths is
+// self-inclusive (the target path is its own last entry, see usePath.js) -
+// dropped here with slice(-1) since revealing a selected row is not a
+// reason to also auto-expand that row's own children: ticking a directory
+// must never expand it, only its caret or name should.
```
pirhoo, line 112: suggestion: small mismatch, the comment says `slice(-1)` but the code is `slice(0, -1)` (and `slice(-1)` would keep only the self entry). The rest reads exactly right.

```suggestion
// Pre-open ancestor directories of selected paths so the tree reveals them
// immediately instead of requiring manual expansion. getAncestorPaths is
// self-inclusive (the target path is its own last entry, see usePath.js) -
// dropped here with slice(0, -1) since revealing a selected row is not a
// reason to also auto-expand that row's own children: ticking a directory
// must never expand it, only its caret or name should.
```

package.json
```diff
@@ -31,14 +31,14 @@
...
-    "@icij/murmur": "5.1.2",
+    "@icij/murmur": "5.1.3",
     "@popperjs/core": "^2.11.8",
     "@tato30/vue-pdf": "^1.11.5",
     "@vueuse/components": "^14.4.0",
     "axios": "^1.13.2",
     "bodybuilder": "^2.5.0",
     "bootstrap": "^5.3.8",
-    "bootstrap-vue-next": "0.42.0",
+    "bootstrap-vue-next": "1.0.1",
```
pirhoo, line 41: question: on BVN 1.0.1 an unchecked `BFormCheckbox` seems to emit `undefined` rather than `false`. Every call site I traced (`exclude`/`contextualize`, `deleteAfterRelaunch`, `all`, the tree checkboxes) is falsy-safe, so probably fine, but did you get a manual pass over those? It's the one behaviour change in the bump I couldn't settle from the tests.

pirhoo (changes requested):
Thanks!

src/components/PathTree/PathTreeView/PathTreeViewEntry.vue
caro3801: Confirmed, done in 3684f41ee. Traced `FormControlPath.vue` (`no-stats`) → used by `ProjectForm`, `TaskDocumentsForm`, `PathBannerForm` — with the stats block gone, there was no navy badge left to carry the "selected" affordance, so removing the whole header-fill rule left selected rows with zero visual indicator there.

Re-added the rule with `--bs-tertiary-bg-subtle` (same token already used for the hover fill a few lines up) instead of `--bs-action`, so it stays a light "selected" tint rather than the reported navy-fill bug, and never needs a forced white-text companion.

src/components/PathTree/PathTreeView/PathTreeViewEntry.vue
caro3801: Confirmed. Applied as-is in bf6de8ebf:

`:active="compactOrInjected || noSearchLink ? selected : active || selected"`

Now `noSearchLink` rows (the path filter, both compact and modal) only go navy on selection, never on hover of a non-clickable span. Added regression tests for all three cases (hover without a search link, selection without a search link, hover with a search link).

src/components/Filter/FilterType/FilterTypePath.vue
caro3801: Fixed in c73e1a2d5, comment now says slice(0, -1).

package.json
caro3801: Did the pass empirically (not just reading the bundle):

- Confirmed the behavior change is real: mounted a bare BFormCheckbox with modelValue: true and unchecked it — update:modelValue emits [undefined], not [false], on 1.0.1.

- exclude/contextualize (FiltersPanelSectionFilterActions.vue) → computedExcludeFilter/computedContextualizeFilter setters → searchStore.toggleFilter(name, toggler) does const excluded = toggler ?? !wasExcluded. ?? treats undefined as "not provided" and falls back to !wasExcluded — which happens to equal the correct new value here only because the checkbox always mirrors the filter's own current state (unchecking always means wasExcluded: true → false, and !wasExcluded gives exactly that). 
Safe in practice, but fragile/coincidental rather than by explicit design — flagging in case you want it hardened later, not blocking this PR.

- deleteAfterRelaunch (BatchSearchActionsRelaunchModal.vue) → read via .value into the submit payload → consumed with if (deleteAfterRelaunch) in BatchSearchActions.vue. Truthy check, falsy-safe.

- "All" checkboxes (FilterTypeAll, FilterTypeProjectAll, PageSettingsSectionGroupAll, DropdownSelectorSelectAll) → all route through computed({ get, set }) setters guarded with if (value). Falsy-safe.

- The path-tree's own selected/indeterminate checkboxes (PathTreeViewEntryNameCheckbox.vue) don't even read the emitted value — PathTree.vue's @update:selected="togglePath(path)" ignores the event payload and re-derives the toggle from isSelectedPath(path). Structurally unaffected by this BVN change.

Every call site checked is falsy/nullish-safe. Nothing to fix for this PR.

pirhoo (approved):

PR #512, opened 2026-08-31
feat: keep the text tab responsive on oversized markdown pages

pirhoo:
Huge markdown structure pages froze the document text tab. This moves rendering off the main thread, caps what gets rendered without consent, and trims the layout cost.

- fix: render markdown structure pages and the markdown viewer in a Web Worker, with an inline fallback when workers are unavailable, a render timeout, and hardened worker failure paths
- fix: skip rendering pages above a size threshold, fall back to plain text, and let the reader render anyway per page from the view dropdown
- fix: keep the rendered page cache across view toggles, evicted per document
- fix: degrade link constraining instead of failing the render on an unresolvable base URL
- perf: mark all search terms in a single parse
- perf: limit content-visibility to tables and code blocks
- build: emit the markdown worker as an ES module chunk

### DIFF
--- a/src/components/Document/DocumentContent.vue
+++ b/src/components/Document/DocumentContent.vue
@@ -323,11 +323,6 @@ function fallbackToTextForEmptyMarkdown() {
   preferMarkdown.value = false
 }
 
-function fallbackToTextForOversizedMarkdown() {
-  markdownOversized.value = true
-  preferMarkdown.value = false
-}
-
 // Both paginations describe the same physical pages when their counts match,
 // so the page number survives the toggle; anything else has no page
 // correspondence and goes back to the first page.
@@ -690,7 +685,7 @@ async function loadContentSliceAround(desiredOffset) {
         :render-oversized="markdownOversized"
         @fallback="preferMarkdown = false"
         @empty="fallbackToTextForEmptyMarkdown"
-        @oversized="fallbackToTextForOversizedMarkdown"
+        @oversized="markdownOversized = true; preferMarkdown = false"
       />
       <div
         v-else-if="hasExtractedContent"
--- a/src/components/Document/DocumentContent.vue
+++ b/src/components/Document/DocumentContent.vue
@@ -325,7 +325,7 @@ function fallbackToTextForEmptyMarkdown() {
 
 // Both paginations describe the same physical pages when their counts match,
 // so the page number survives the toggle; anything else has no page
-// correspondence and goes back to the first page.
+// correspondence, so each side keeps its own last position.
 async function syncPagePosition(markdown) {
   // A translation forces text mode through its own contract (start at offset
   // 0, handled by the targetLanguage watcher): its offsets describe another
@@ -336,7 +336,12 @@ async function syncPagePosition(markdown) {
   }
   const aligned = !!syncedPages.value?.length && syncedPages.value.length === markdownPagesCount.value
   if (markdown) {
-    markdownPage.value = aligned ? pageForOffset(activeContentSliceOffset.value) : 1
+    // An unaligned flip is a mode toggle, not a navigation: a page-1 reset here
+    // would fire the `markdownPage` watcher and silently clear the "render
+    // anyway" consent the reader just gave for the page they were on.
+    if (aligned) {
+      markdownPage.value = pageForOffset(activeContentSliceOffset.value)
+    }
     return
   }
   // The restored page needs its text slice loaded, and going through
@@ -686,6 +691,7 @@ async function loadContentSliceAround(desiredOffset) {
         @fallback="preferMarkdown = false"
         @empty="fallbackToTextForEmptyMarkdown"
         @oversized="markdownOversized = true; preferMarkdown = false"
+        @rendered="markdownOversized = false"
       />
       <div
         v-else-if="hasExtractedContent"
--- a/src/components/Document/DocumentContent.vue
+++ b/src/components/Document/DocumentContent.vue
@@ -64,6 +64,10 @@ const preferMarkdown = ref(true)
 // That only shows once its first page comes back empty, so the toggle is built
 // from the manifest and this runtime finding together.
 const isMarkdownEmpty = ref(false)
+// An oversized page already sent the reader to plain text once for this
+// document: the dropdown's "Formatted" entry then becomes their explicit
+// "render anyway", so the child renders it instead of re-emitting.
+const markdownOversized = ref(false)
 const markdownPage = ref(1)
 const markdownMatches = ref([])
 // The term the occurrence counts were computed for. The marks a page renders
@@ -251,6 +255,7 @@ watch(isMarkdownMode, async (markdown) => {
 watch(docId, async () => {
   preferMarkdown.value = true
   isMarkdownEmpty.value = false
+  markdownOversized.value = false
   markdownPage.value = 1
   markdownMatches.value = []
   markdownAppliedTerm.value = ''
@@ -309,6 +314,11 @@ function fallbackToTextForEmptyMarkdown() {
     return
   }
   isMarkdownEmpty.value = true
+  preferMarkdown.value = false
+}
+
+function fallbackToTextForOversizedMarkdown() {
+  markdownOversized.value = true
   preferMarkdown.value = false
 }
 
@@ -638,6 +648,7 @@ async function loadContentSliceAround(desiredOffset) {
           v-if="hasMarkdown"
           v-model="preferMarkdown"
           :markdown-disabled="isMarkdownEmpty"
+          :markdown-slow="markdownOversized"
           :translation="isTranslation"
           class="flex-shrink-0 ms-auto"
         />
@@ -670,8 +681,10 @@ async function loadContentSliceAround(desiredOffset) {
         :term="markdownAppliedTerm"
         :global-search-terms="globalSearchTerms"
         :active-match="activeMarkdownMatch"
+        :render-oversized="markdownOversized"
         @fallback="preferMarkdown = false"
         @empty="fallbackToTextForEmptyMarkdown"
+        @oversized="fallbackToTextForOversizedMarkdown"
       />
       <div
         v-else-if="hasExtractedContent"
--- a/src/components/Document/DocumentContent.vue
+++ b/src/components/Document/DocumentContent.vue
@@ -249,6 +249,12 @@ watch(isMarkdownMode, async (markdown) => {
   }
 })
 
+// Consent to render an oversized page is given for that page, not for the
+// document: the next page gets the size guard again.
+watch(markdownPage, () => {
+  markdownOversized.value = false
+})
+
 // The manifest, the page and the matches all describe one document. The mount
 // probe cannot cover a host that swaps the prop without remounting, so the
 // document identity re-runs it and clears what belonged to the previous one.
--- a/src/components/Document/DocumentContentMarkdown.vue
+++ b/src/components/Document/DocumentContentMarkdown.vue
@@ -2,7 +2,7 @@
 import { computed, nextTick, reactive, ref, toRef, useTemplateRef, watch } from 'vue'
 import { useI18n } from 'vue-i18n'
 
-import { addSearchMarksClassInHtml } from '@/utils/strings'
+import { addSearchMarksClassesInHtml } from '@/utils/strings'
 import { renderMarkdownOffThread } from '@/utils/markdownOffThread'
 import { useMarkdownAnchors } from '@/composables/useMarkdownAnchors'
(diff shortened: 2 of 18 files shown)


src/components/Document/DocumentContent.vue
caro3801, line 337: crucial: Looks like toggling back to Formatted on an unaligned document resets `markdownPage` to 1,
which silently clears the reader's just-given "render anyway" consent through the `markdownPage`
watcher below, before they ever see the page that triggered it.

Since `docId`'s watcher already resets `markdownPage` to 1 explicitly for a real document change,
could this branch leave the current page alone when the flip is just a mode toggle rather than a
new document, so the reader lands back on the oversized page they consented to render?

src/components/Document/DocumentContent.vue
```diff
@@ -245,12 +249,19 @@ watch(isMarkdownMode, async (markdown) => {
   }
 })
 
+// Consent to render an oversized page is given for that page, not for the
+// document: the next page gets the size guard again.
+watch(markdownPage, () => {
+  markdownOversized.value = false
+})
```
caro3801, line 256: important: Could we also clear `markdownOversized` once the page actually renders successfully, not
only on a page/docId change? For a single-page markdown document that trips the guard once, neither
watcher fires again, so the "may be slow" warning stays on the dropdown forever even though
`renderPageOnce`'s cache hit now serves that page instantly.

A tidy option would be having `DocumentContentMarkdown` emit something like `rendered` once
`renderPageOnce` returns `false` (successfully rendered, not oversized), and clearing the flag from
that instead of only from the page/doc watchers.

src/components/Document/DocumentContentMarkdown.vue
```diff
@@ -139,11 +246,22 @@ async function renderPageOnce() {
...
   }
-  const { index, id, routing } = props.document
-  const markdown = await api.getStructurePage(index, id, targetPage, routing)
-  renderedPages[targetCacheKey] = await renderMarkdown(markdown)
+  const markdown = await fetchPageSource(targetCacheKey, targetPage)
+  if ((markdown?.length ?? 0) > props.oversizedThreshold && !props.renderOversized) {
+    oversizedSources[targetCacheKey] = markdown
+    return true
+  }
+  const html = await renderMarkdownOffThread(markdown)
```
caro3801, line 258: with the fix above https://github.com/ICIJ/datashare-client/pull/512/changes#r3896355109
```suggestion
  const html = await renderMarkdownOffThread(markdown)
  delete oversizedSources[targetCacheKey]
```

src/components/Document/DocumentContentMarkdown.vue
```diff
@@ -130,7 +227,17 @@ function reportEmptyPage() {
...
-async function renderPageOnce() {
+async function fetchPageSource(cacheKey, page) {
+  if (cacheKey in oversizedSources) {
+    const source = oversizedSources[cacheKey]
+    delete oversizedSources[cacheKey]
+    return source
+  }
+  const { index, id, routing } = props.document
+  return api.getStructurePage(index, id, page, routing)
+}
```
caro3801, line 240: crucial: Looks like `fetchPageSource` deletes the cached raw markdown from `oversizedSources` as soon
as it is read, before `renderMarkdownOffThread` has actually succeeded. If the render then throws
(worker timeout, worker death, `messageerror`), the already-downloaded oversized payload is gone from both caches, so hitting Retry re-downloads the whole thing again, right on the path (a huge page)
where avoiding that was the point of the cache.

Could the delete move to `renderPageOnce`, right after `renderMarkdownOffThread` resolves, instead of inside `fetchPageSource`?


```suggestion
async function fetchPageSource(cacheKey, page) {
  if (cacheKey in oversizedSources) {
    return oversizedSources[cacheKey]
  }
  const { index, id, routing } = props.document
  return api.getStructurePage(index, id, page, routing)
}
```
with the addition fix below

src/utils/markdownOffThread.js
```diff
@@ -0,0 +1,99 @@
...
+  if (!source) {
+    return Promise.resolve('')
+  }
+  if (typeof Worker === 'undefined') {
+    return renderInline(source)
+  }
+  const id = ++lastRenderId
+  const base = window.location.href
+  return new Promise((resolve, reject) => {
+    const timer = setTimeout(() => failWorker('Markdown worker timed out'), RENDER_TIMEOUT)
```
caro3801: crucial: A single render's timeout currently calls `failWorker`, which terminates the shared worker
and rejects every other in-flight render sharing it, even ones for a completely unrelated document
that were rendering fine. Could the timeout just settle its own entry instead of tearing down the
whole worker and queue?

```suggestion
    const timer = setTimeout(() => settle(id)?.reject(new Error('Markdown worker timed out')), RENDER_TIMEOUT)
```

src/components/Document/DocumentViewer/DocumentViewerMarkdown.vue
caro3801, line 58: suggestion: This component has no staleness guard around the fetch-plus-off-thread-render pipeline,
unlike the `lastPageLoad` pattern this PR adds to `DocumentContentMarkdown.vue`. Could we bring the same guard here? 
Right now, switching quickly from a large, slow-rendering document to a small, fast one can let the large one's render resolve last and overwrite `html` with stale content for a document that is no longer selected. 
This could go as a follow up if you think it's interesting to add. If this pattern appears at multiple places, it could also become a composable

```suggestion
let lastLoad = 0

async function load(document) {
  if (!document) {
    return
  }
  const id = ++lastLoad
  loading.value = true
  error.value = null
  try {
    const source = await fetchSource(document, { responseType: 'text' })
    const rendered = await renderMarkdownOffThread(source)
    if (id !== lastLoad) {
      return
    }
    html.value = rendered
  }
  catch (e) {
    if (id === lastLoad) {
      error.value = e.message
    }
  }
  finally {
    if (id === lastLoad) {
      loading.value = false
    }
  }
}
```

src/components/Document/DocumentContentMarkdown.vue
```diff
@@ -251,3 +367,16 @@ watch(toRef(props, 'activeMatch'), activateMatch, { flush: 'post' })
...
   </div>
 </template>
+
+<style lang="scss">
+.document-content-markdown__body {
+  // Tables and code blocks are the expensive blocks, so only they skip layout
+  // and paint while off screen. Paragraphs lay out normally, which keeps the
+  // scroll offsets an anchor jump or a mark's scrollIntoView lands on accurate.
+  > table,
+  > pre {
```
caro3801: suggestion: The `content-visibility: auto` selector is scoped to `> table, > pre` (direct children
only), so a table or code block nested inside a blockquote or a list item, which this repo's own
fixtures already exercise, gets none of the layout/paint deferral this PR adds. 

Could the child combinator be dropped so nested blocks benefit too?

```suggestion
  table,
  pre {
```

src/components/Document/DocumentContentMarkdown.vue
```diff
@@ -1,13 +1,68 @@
...
+  dropUnshownDocuments()
+}
+
+function hideDocument(documentId) {
+  const count = shownDocumentIds.get(documentId)
+  if (count === undefined) {
+    return
+  }
+  shownDocumentIds.set(documentId, count - 1)
+}
```
caro3801, line 52: suggestion: The module-scope `renderedPages`/`oversizedSources` caches plus the hand-rolled
`shownDocumentIds`/`showDocument`/`hideDocument`/`dropUnshownDocuments` reference counting solve the problem Vue's built-in `<KeepAlive>` already solves (surviving the Formatted/Plain
`v-if` unmount). 
This component only has one call site, so the cross-instance sharing this ~50 lines defends against does not happen today. Wrapping the toggle in `DocumentContent.vue` with `<KeepAlive>`
would get the same caching for free with no bookkeeping to keep correct. 
Worth a follow-up rather than blocking this PR on it.

src/components/Document/DocumentContentMarkdown.vue
caro3801, line 302: question: `activateMatch`'s `scrollIntoView` (and the anchor-jump in `useMarkdownAnchors`) can target
a mark or heading inside a table/pre that `content-visibility: auto` is currently skipping. 
Do we know how the browsers we support resolve `scrollIntoView` against a `content-visibility: auto`
ancestor here, that is, does it land on the placeholder's estimated 300px height or the block's real
size once expanded? 
If it is the placeholder, matches inside large off-screen tables could land in the wrong spot.

caro3801 (changes requested):
super nice! getting the oversized-markdown path responsive, the worker offload and the staleness counters is a real gap in perf for markdown :). There are two crucial cache/consent bugs (https://github.com/ICIJ/datashare-client/pull/512/changes#r3896355109 and https://github.com/ICIJ/datashare-client/pull/512/changes#r3896367460) that are worth fixing before merge since they undo the exact caching and worker-sharing your PR sets out to add.
The rest are smaller robustness gaps that can land as follow-ups without holding this up.

src/components/Document/DocumentContentMarkdown.vue
pirhoo: Good catch, that delete was exactly on the wrong side of the render. Moved it to `renderPageOnce`, right after the write to `renderedPages`, so a failed render keeps the payload and Retry serves it from cache. Fixed in 663b10444, with a test covering the failing-render-then-retry path.

src/components/Document/DocumentContentMarkdown.vue
pirhoo: Applied in 663b10444, with one nuance: the delete sits after the staleness check rather than right after the render, so a superseded load also keeps the payload around for the load that replaced it.

src/utils/markdownOffThread.js
pirhoo: Agreed on the diagnosis, but the suggested one-liner has a downside: the wedged worker stays alive, so every future render queues behind it and times out too, and markdown rendering is dead until reload. Since the worker answers in posting order, the first timer to fire always belongs to the render the worker is stuck on. So in e49d82c86 the timeout now fails only its own render, terminates the stuck worker, and reposts the queued renders to a fresh one. Innocent renders survive and the worker never wedges permanently.

src/components/Document/DocumentContent.vue
pirhoo: Right, the page-1 reset was silently eating the consent through the watcher. `syncPagePosition` now leaves the page alone on an unaligned flip, so the reader lands back on the page they just agreed to render. The `docId` watcher still resets to 1 for a real document change. Fixed in 663b10444 with a test.

src/components/Document/DocumentContent.vue
pirhoo: Went with your `rendered` event suggestion: `DocumentContentMarkdown` emits it once `loadPage` completes without tripping the guard (cache hits included), and the parent clears `markdownOversized` from that. Fixed in 663b10444.

src/components/Document/DocumentViewer/DocumentViewerMarkdown.vue
pirhoo: Added your guard in 99129123d, same counter pattern as `DocumentContentMarkdown`, plus a test for the slow-then-fast document swap. Held off on the composable for now since that would make two call sites with slightly different shapes; happy to extract it the day a third one shows up.

src/components/Document/DocumentContentMarkdown.vue
pirhoo: Dropped the combinator in 876bb5e02, nested tables and code blocks now get the same deferral.

src/components/Document/DocumentContentMarkdown.vue
pirhoo: Fair point, `<KeepAlive>` would delete most of that bookkeeping. The one thing it does not give for free is the current eviction contract (pages cached per document and dropped when nobody shows the document anymore), so I would rather try that swap in its own PR where we can compare memory behavior. Keeping it as a follow-up.

src/components/Document/DocumentContentMarkdown.vue
pirhoo: Good question. `scrollIntoView` forces layout of the target and its ancestors, so it does not land on the placeholder itself. The residual risk is the skipped blocks *before* the target: their 300px estimate is what positions everything below them, so the landing can drift until they expand. Two things bound this in practice: `contain-intrinsic-size: auto` remembers the real size once a block has rendered, and marks/anchors inside tables are rare compared to prose. I will keep an eye on it with the big-table fixtures and revisit if jumps land visibly off.

caro3801 (approved):
Thanks!

